Issue #34, opened 2020-02-20
[bug] [qemu] arm64/aarch64 missing .dylib


Marcinoo97:
UTM claims to support 30 processor architectures but arm64/aarch64 refused to start when I tried. 

![IMG_20200220_1837382](https://user-images.githubusercontent.com/2014596/74967511-08616800-5411-11ea-8162-0eb4d29c9893.jpg)

iosdec:
@Marcinoo97 have you got os's working so far? if so could you provide a link to a working one and the setup you're using? not having any luck so far with anything

ledwardchow:
It's loading the dylib filename from systemArchitecturesPretty instead of systemArchitectures.

x86 works because it's the same name by coincidence.

Replace line 158 of VMConfigSystemViewController.m:
        self.configuration.systemArchitecture = self.architectureLabel.text;
with:
        self.configuration.systemArchitecture = [UTMConfiguration supportedArchitectures][row];

osy86:
Merged #37

PR #37, opened 2020-02-21
aarch64 fixes:

ledwardchow:
* launch qemu with -machine virt
* remove qxl
* fixed dylib filename generation


ledwardchow:
fixes #34

osy86:
Does `virtio-gpu-pci` actually work? I don’t think it’s actually supported. Would be better to just drop the vga option and use the default.

pixelomer:
This also fixed PowerPC for me.

Atemu:
@ledwardchow why was qxl disabled?

Is the reason still relevant with modern versions of qemu?

Issue #32, opened 2020-02-20
[UTM] App freezes/white screen on launch


unorderedpizza:
iOS 13.3 on iPhone XS.

While not jailbroken, the app itself can be opened as long as it is signed, however, the app will crash right after clicking on any of the VMs. Configuration menu works without issues.

Additionally, the app will crash when a fairly large VM is attempting to be launched even in jailbroken mode (in my case, a Windows XP install with a 2 GB hard drive, it was set up on the device itself and was fully functional until I shut down the VM and attempted to re-launch the virtual machine later on); when reopened, the app will stay on the white splash screen after the initial crash and stop functioning, even after removing the app in the switcher/respringing/su && ldrestart (possibly due to lack of cleanup?). This is also reproducible by shutting down the virtual machine with the power button on the top menu UI; the machine shuts down and the app freezes. Only fixable by rebooting and re-jailbreaking.
![IMG_0660](https://user-images.githubusercontent.com/43310440/74922700-f41e6a80-5412-11ea-8c83-c74f3eadd8d3.PNG)

Tizioincognit0:
I have the same issue on iOS 12.2 on iPhone X.

tylerdak:
Similar issue here, iOS 13.3, iPhone XS

obbcth:
Same issue on iOS 13.3, iPad Air 3, Jailbroken.

JohnnnnyKlayy:
Same issue, iOS 13.3 iPhone XR, jailbroken.

osy86:
This is a known issue and is due to an iOS bug. If an app dies (exception, killed, etc) with `ptrace` active, then it will lock up the kernel (I think).

obbcth:
> This is a known issue and is due to an iOS bug. If an app dies (exception, killed, etc) with `ptrace` active, then it will lock up the kernel (I think).

Same as ppsspp jit debug ver?
Then there should be a app close button to halt ptrace, I think.

seriousm4x:
Same issue on iPad Pro 3 12,9" on iOS 13.3 jailbroken

realmayus:
Same issue on iPad Pro 12.9“ 2018 on iOS 13.3.1 non-jailbroken

The-SamminAter:
Same issue, iOS 13.2.3 on an iPhone XS

saagarjha:
I believe there are ways around this. I’ll make a PR if I can figure this out.

Wthebinder:
[...]

PR #38, opened 2020-02-21
Improve usability

saagarjha:
A somewhat trivial pull request to make working with the project a bit easier as I get acquainted with the system. Finally got it to build…

(FYI: it's `PT_TRACE_ME` on Darwin; that was an oversight on my part when I left that comment)

Update: this also fixes another big annoyance, #32.

### DIFF
--- a/UTM/main.m
+++ b/UTM/main.m
@@ -15,31 +15,63 @@
 //
 
 #import <UIKit/UIKit.h>
+#import <mach/mach.h>
+#import <pthread.h>
 #import "AppDelegate.h"
 
+extern boolean_t exc_server(mach_msg_header_t *, mach_msg_header_t *);
 extern int ptrace(int request, pid_t pid, caddr_t addr, int data);
 
 #define PT_TRACE_ME 0
-#define PT_DENY_ATTACH 31
+#define PT_SIGEXC 12
+
+kern_return_t catch_exception_raise(mach_port_t exception_port,
+                                    mach_port_t thread,
+                                    mach_port_t task,
+                                    exception_type_t exception,
+                                    exception_data_t code,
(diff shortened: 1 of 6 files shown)


UTM/main.m
```diff
@@ -43,7 +44,16 @@ kern_return_t catch_exception_raise(mach_port_t exception_port,
     return NULL;
 }
 
+static BOOL am_i_being_debugged() {
```
coolstar: better to use csops to check for CS_DEBUGGED here, in case it was set by a jailbreak already (as jailbreaks can set the codesigning flag without having to call ptrace)

saagarjha: I don’t see why this case is relevant, as I would expect a version designed for jailbroken devices to be signed with dynamic-codesigning rather than doing tricks with `CS_DEBUGGED`…

coolstar: 2 things about the dynamic-codesigning entitlement

To actually use it, the app needs to use mmap with MAP_JIT, and it needs to be inside a sandbox.

Without MAP_JIT, you’re basically using ptrace/CS_DEBUGGED

saagarjha: I’m aware of those requirements; I don’t see either of them being an issue in this case?

saagarjha: On that note, I should probably send a fix for https://github.com/hrydgard/ppsspp/issues/9186 to the PPSSPP people as well.

NyaMisty:
Sorry could you please upload an ipa for us to test? Thanks a lot

saagarjha:
GitHub wouldn't take it, so I'll self host it temporarily: http://server.saagarjha.com/UTM.ipa

PR #61, opened 2020-02-24
Add i18n for zh-Hans

Evsio0n:


Evsio0n:
Also add Fix tabbar View when tapping "CPU count" and "Trying first boot from..."

PR #70, opened 2020-02-24
Assume user is quitting and terminate in switcher

unorderedpizza:
This will prevent #32, but disallow backgrounding


osy86:
This is a bad idea. If you get a phone call it kills the vm.

saagarjha:
Right, this is why I only exited when the app was going to terminate.

PR #96, opened 2020-03-03
Fix incomplete parameters

js-john:
1. Fix incomplete parameters for cpu setting.
2. Fix translate error "z系统" to "系统"
3. perform a network requrest in order to show up the alert of "allow network usage"

### DIFF
--- a/Managers/UTMQemuSystem.m
+++ b/Managers/UTMQemuSystem.m
@@ -36,7 +36,7 @@
     [self pushArgv:@"-qmp"];
     [self pushArgv:@"tcp:localhost:4444,server,nowait"];
     [self pushArgv:@"-smp"];
-    [self pushArgv:[NSString stringWithFormat:@"cpus=%@", self.configuration.systemCPUCount]];
+    [self pushArgv:[NSString stringWithFormat:@"cpus=%@,sockets=1", self.configuration.systemCPUCount]];
     if ([self.configuration.systemArchitecture isEqualToString:@"aarch64"]) {
         [self pushArgv:@"-machine"];
         [self pushArgv:@"virt"];
--- a/UTM/zh-Hans.lproj/Main.strings
+++ b/UTM/zh-Hans.lproj/Main.strings
@@ -362,7 +362,7 @@
 "rbw-FB-I5Q.text" = "hda1";
 
 /* Class = "UILabel"; text = "System"; ObjectID = "riB-R9-dTb"; */
-"riB-R9-dTb.text" = "z系统";
+"riB-R9-dTb.text" = "系统";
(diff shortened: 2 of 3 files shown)


osy86:
Why is “sockets=1” needed? What error does this fix?

js-john:
> Why is “sockets=1” needed? What error does this fix?

I tried if the cpu setting in "Architecture" page is more than 2 cores, the setting will not take effect. If the parameter "socket = 1" is included, you can set more than 2 cores. Check it in the task manager.

osy86:
I can not reproduce the “allow network usage" alert. How do you get it to show up? Does setting https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsallowsarbitraryloads?language=objc do anything?

js-john:
> I can not reproduce the “allow network usage" alert. How do you get it to show up? Does setting https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsallowsarbitraryloads?language=objc do anything?

I am not sure whether the "Allows arbitrary loads" still works...

To reproduce the “allow network usage" alert:
Change a new bundle identifier, build,  and run UTM in your iPhone. When you performing a network request (not in the VMs) the "Allow UTM to use wireless data?" alert will show up.

If this alert did not shown, UTM will not have network usage permission. I tried on my phone, even if DNS is configured correctly, I can't access the network in Internet Explorer and the "Allow UTM to use wireless data?" alert did not show up...

I think it is necessary to add network request codes somewhere in the project...

network

osy86:
Are you sure that is from iOS and not some jailbreak tweak? I can't reproduce it on my phone.

js-john:
> Are you sure that is from iOS and not some jailbreak tweak? I can't reproduce it on my phone.

I'm sure this is from iOS. Since maybe iOS 12? or 13. My phone is never jailbroken.

js-john:
> Are you sure that is from iOS and not some jailbreak tweak? I can't reproduce it on my phone.

I am using iOS13.3.1, all apps have to get network access through this alert

saagarjha:
A China thing, I belive.

js-john:
Ok, I see.

osy86:
That makes sense, I think I want the way ish does it because it looks cleaner and I think the Apple url is smaller in data usage.

js-john:
Thank you!

osy86:
Merged https://github.com/utmapp/UTM/commit/6d3f7d049f7a20282065c0a7f2608db7acb70140 https://github.com/utmapp/UTM/commit/c475ad7975676f24675f71ff780487260b83a145

Issue #93, opened 2020-03-03
[Configuration] Fix additional qemu args


osy86:
We currently don't seem to handle additional args correctly. We should provide the user with a way to add extra arguments in a list. We should also display the full set of args for debugging purposes.

PR #108, opened 2020-03-06
Implement argument list view

spotlightishere:
This partially completes #93. It provides an additional table view separating boot arguments, and has a working method to split the list argument into valid, passable arguments.

Some UX is rather janky in regards to removing inputted text from the UI, so I am submitting this as-is via draft. It additionally does not display flags visually anywhere as the issue calls for.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,4 +68,4 @@ fastlane/test_output
 iOSInjectionProject/
 
 build/
-sysroot/
+sysroot*/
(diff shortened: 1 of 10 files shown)


saagarjha:
Screenshots would be nice ;)

spotlightishere:
Of course, thank you for the reminder!

On iPhone:
iPhone - System screen iPhone - main arg screen iPhone - deletion screen

On iPad, to show scaling purposes:
iPad - main arg screen

osy86:
Looks good so far

osy86:
> It additionally does not display flags visually anywhere as the issue calls for.

I'm fine with this not being implemented because I am about to implement logging and the args will always be in the logs. Otherwise, it looks like this is ready to be reviewed?

spotlightishere:
Sounds good to me - as far as I’ve tested alone, this alone should be final

Issue #97, opened 2020-03-04
The translation of simplified Chinese is not accurate enough


WindowsBetaTest:
Some of the warnings are still in English.
Some Chinese characters are followed by some messy English letters.

tihmstar:
Can you provide some concrete examples please?
Like screenshots, steps to reproduce or maybe point to the piece of code responsible for it

PR #114, opened 2020-03-08
Audio and full localization for zh-Hans

js-john:
#4 
Test audio successfully in XP, Windows 7, Windows 10
#97
Localization of messages, warnings, operations has been completed


js-john:
Adding the "No more warning" option to memory warnings may have a better experience.

PR #142, opened 2020-03-15
Redirect stdout/stderr to log file

### LINKED ISSUE #99: [qemu] Capture log to file
Redirect stdout and stderr to a log file that we keep in the .utm. Add a setting option to share that file. Would improve debugging.

osy86:
This adds a new option to log to file and a way to export the log for debug purposes.

Resolves #99

### DIFF
--- a/Managers/UTMLogging.m
+++ b/Managers/UTMLogging.m
@@ -1,0 +1,215 @@
+//
+// Copyright © 2020 osy. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


Managers/UTMLogging.m
```diff
@@ -0,0 +1,215 @@
...
+            NSLog(@"Trying to init more than one instance of UTMLogging!");
+            return nil;
+        }
+        int success = 0;
+        int new_stdout[2];
+        int new_stderr[2];
+        _stdout = _stderr = -1;
+        _real_stdout = _real_stderr = -1;
+        new_stdout[0] = new_stdout[1] = -1;
+        new_stderr[0] = new_stderr[1] = -1;
```
saagarjha:
```
int new_stdout[] = {-1, -1};
int new_stderr[] = {-1, -1};
```?

osy86: Oops sorry didn't see this comment! I thought of that but decided it's more consistent with the other lines of code to have it this way. It clarifies the intent that those different variables are related to each other.

saagarjha: Fair enough. I think you actually committed first, and I posted my comment before I got sent a notification ;)

PR #174, opened 2020-03-24
Sound card configurable. Localizable for zh-Hans.

js-john:
IMG_3944
IMG_3945


Configuration/UTMConfiguration.m
```diff
@@ -1327,6 +1340,7 @@ - (id)initDefaults:(NSString *)name {
         self.networkEnabled = YES;
         self.printEnabled = YES;
         self.soundEnabled = YES;
+        self.soundCardDeviceId = @0;
```
osy86: For other values such as architectures and targets, we don't store the index but the actual value. Please do this as well. We don't want to force a mapping of id to value for future compatibility.

osy86:
Please `git rebase` to master instead of merging master.

js-john:
ok

osy86:
The PR is good, you just need to make some changes.

PR #183, opened 2020-03-27
Add support for aarch64

lshzh-ww:
Unlike other architectures, QEMU doesn’t provide a bios file for aarch64 by default. This fix manually sets a EFI firmware for aarch64 VMs and improves some configurations.


lshzh-ww:
should change

Issue #160, opened 2020-03-21
Improved pointer support on iOS 13.4


saagarjha:
Since iOS 13.4 has improved cursor support, perhaps we should rethink some of the interactions if the user has a mouse or trackpad connected. For example, hiding the cursor and disabling some of the gestures (as they are now handled by the hardware) might be a good idea.

osy86:
Is there any API to detect if a mouse is connected?

saagarjha:
I don't think it's dropped yet. I wonder when the next Xcode beta will be out…

itselijahciali:
I know that it's supposed to snap to buttons, so theoretically somebody could make a program to detect the Windows cursor status, and report it to the app, and the app places a control there. Of course, we could just wait for an API...

osy86:
FYI #176 will track work supporting cursor inside the VM. This issue is for better handling for trackpad to navigate UTM.

conath:
I implemented cursor handling in the VM list view, VM metal view toolbar and (for #176) the actual cursor movement. :)

Issue #122, opened 2020-03-09
Touchscreen input option does not work


Shawnjoseph2001:
The option to use a touchscreen as the input device in Android x86 4.3 and possibly other OS's does not work. A workaround is to use the trackpad input option.

osy86:
I think touchscreen input will be a future option. But it requires client side support. I don't know if android x86 has it.

cs-sugar:
There should be.

Shawnjoseph2001:
If it does not work, the touchscreen support should be labeled as experimental or non-working within the app.

osy86:
Sir the app is labeled as experimental.

cs-sugar:
Emm. But I’m not a Sir.


| |
彩虹小糖
|
|
邮箱：abcd2890000456@126.com
|

签名由 网易邮箱大师 定制

On 03/10/2020 10:52, osy86 wrote:

Sir the app is labeled as experimental.

osy86:
Replying to OP.

ThatsNiceGuy:
The touchscreen input option does not work as expected in windows 10 with the SPICE drivers installed. I touch on screen and the input that the OS receives is completely inaccurate. Can provide a video if needed, but i feel this is an issue that should be fixed

PR #193, opened 2020-03-28
Implement iPadOS 13.4 Pointer Interaction for UI and VM

### LINKED ISSUE #176: iPadOS 13.4 pointer support
https://developer.apple.com/documentation/uikit/pointer_interactions?changes=latest_minor&language=objc

conath:
This PR fixes #176 and contains parts of #160 

Adds pointer interaction support to:
UI:
- VM List cells
- VM View toolbar buttons

VM:
- Cursor movement
- Auto hiding iPadOS cursor when inside VM display area

Missing:
- Configuration views, any other views not mentioned

### DIFF
--- a/Views/VMDisplayMetalViewController+Pointer.m
+++ b/Views/VMDisplayMetalViewController+Pointer.m
@@ -1,0 +1,85 @@
+//
+// Copyright © 2020 osy. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import "VMDisplayMetalViewController.h"
+#import "VMDisplayMetalViewController+Touch.h"
+#import "VMDisplayMetalViewController+Pointer.h"
+#import "VMCursor.h"
+#import "CSDisplayMetal.h"
+
+NS_AVAILABLE_IOS(13.4)
+@implementation VMDisplayMetalViewController (Pointer)
+
+// Add pointer interaction to VM view
+-(void)initPointerInteraction {
+    [self.mtkView addInteraction:[[UIPointerInteraction alloc] initWithDelegate:self]];
+}
+
+#pragma mark - UIPointerInteractionDelegate
+- (UIPointerStyle *)pointerInteraction:(UIPointerInteraction *)interaction styleForRegion:(UIPointerRegion *)region {
+    // Hide cursor while hovering in VM view
+    if (interaction.view == self.mtkView) {
+        return [UIPointerStyle hiddenPointerStyle];
+    }
+    return nil;
+}
+
+static CGFloat CGPointToPixel(CGFloat point) {
+    return point * [UIScreen mainScreen].scale; // FIXME: multiple screens?
+}
+
+- (bool)isPointOnVMDisplay:(CGPoint)pos {
+    CGSize screenSize = self.mtkView.drawableSize;
+    CGSize scaledSize = {
+        self.vmDisplay.displaySize.width * self.vmDisplay.viewportScale,
+        self.vmDisplay.displaySize.height * self.vmDisplay.viewportScale
+    };
+    CGRect drawRect = CGRectMake(
+        self.vmDisplay.viewportOrigin.x + screenSize.width/2 - scaledSize.width/2,
+        self.vmDisplay.viewportOrigin.y + screenSize.height/2 - scaledSize.height/2,
+        scaledSize.width,
+        scaledSize.height
+    );
+    pos.x -= drawRect.origin.x;
+    pos.y -= drawRect.origin.y;
+    return 0 <= pos.x && pos.x <= scaledSize.width && 0 <= pos.y && pos.y <= scaledSize.height;
+}
+
+
+- (UIPointerRegion *)pointerInteraction:(UIPointerInteraction *)interaction regionForRequest:(UIPointerRegionRequest *)request defaultRegion:(UIPointerRegion *)defaultRegion {
+    // Requesting region for the VM display?
+    if (interaction.view == self.mtkView) {
+        // Then we need to find out if the pointer is in the actual display area or outside
+        CGPoint location = [request location];
+        CGPoint translated = location;
+        translated.x = CGPointToPixel(translated.x);
+        translated.y = CGPointToPixel(translated.y);
+        
+        if ([self isPointOnVMDisplay:translated]) {
+            // move vm cursor, hide iOS cursor
+            CGPoint clipped = [self clipCursorToDisplay:translated];
+            [_cursor setCenter:clipped];
+            return [UIPointerRegion regionWithRect:[self.mtkView bounds] identifier:@"vm view"];
+        } else {
+            // don't move vm cursor, show iOS cursor
+            return nil;
+        }
+    } else {
+        return nil;
+    }
+}
+
+@end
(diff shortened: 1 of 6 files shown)


osy86:
Is there any reason why we need to manually code cursor support for UIButton? In #192 I just set the `pointerInteraction` attribute in the storyboard.

conath:
> Is there any reason why we need to manually code cursor support for UIButton? In #192 I just set the `pointerInteraction` attribute in the storyboard.

Good catch, I hadn't thought of doing it in the storyboard. Just tried it. That unfortunately results in the "preview" effect applied to the button, which is incompatible with the Visual effect view (causes the button to render completely black).

It seems like the CI that auto builds here is still on Xcode 13.3.1, therefore fails to build the project with the reference to new API. Not sure if I can fix that on my end?

osy86:
> That unfortunately results in the "preview" effect applied to the button, which is incompatible with the Visual effect view (causes the button to render completely black).

Right, that's why I also changed the contrast so the buttons are white (which frankly I like more). I think I might merge that and then rework this so it just handles the cursor in the VM. What do you think?

> It seems like the CI that auto builds here is still on Xcode 13.3.1, therefore fails to build the project with the reference to new API. Not sure if I can fix that on my end?

That's fine.

conath:
> Right, that's why I also changed the contrast so the buttons are white (which frankly I like more). I think I might merge that and then rework this so it just handles the cursor in the VM. What do you think?

Sure! I can do that if you want, it's just a matter of rolling back some of the changes I made. Should I keep the effect for the vm list? It magnifies the cell slightly on hover.

Views/VMDisplayMetalViewController.m
```diff
@@ -101,6 +102,13 @@ - (void)viewDidLoad {
     self.mtkView.delegate = _renderer;
     
     [self initTouch];
+    // Pointing device support on iPadOS 13.4 GM or later
+    if (@available(iOS 13.4, *)) {
+        // Betas of iPadOS 13.4 did not include this API, that's why I check if the class exists
+        if (NSClassFromString(@"UIPointerInteraction") != nil) {
```
saagarjha: Can you weak-link it?

Views/VMDisplayMetalViewController+Pointer.m
```diff
@@ -0,0 +1,116 @@
...
+            CGPoint clipped = [self clipCursorToDisplay:translated];
+            [_cursor setCenter:clipped];
+            return [UIPointerRegion regionWithRect:[self.mtkView bounds] identifier:@"vm view"];
+        } else {
+            // don't move vm cursor, show iOS cursor
+            return nil;
+        }
+    } else {
+        // Is it a UIControl (e. g. Button)?
+        if ([interaction.view.class respondsToSelector:NSSelectorFromString(@"isEnabled")]) {
```
saagarjha: Are you sure this is the right check? From a cursory glance it seems like you're using the wrong receiver…also, can you not just use `@selector`?

Views/VMDisplayMetalViewController+Pointer.m
```diff
@@ -0,0 +1,116 @@
...
+        return false;
+    } else if (pos.x > scaledSize.width) {
+        return false;
+    }
+    if (pos.y < 0) {
+        return false;
+    } else if (pos.y > scaledSize.height) {
+        return false;
+    }
+    return true;
```
saagarjha: If not the above, `return 0 <= pos.x && pos.x <= scaledSize.width && 0 <= pos.y && pos.y <= scaledSize.height`?

Views/VMDisplayMetalViewController+Pointer.m
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+        return [UIPointerStyle styleWithShape:[UIPointerShape shapeWithRoundedRect:interaction.view.bounds] constrainedAxes:UIAxisBoth];
+    }
+}
+
+static CGFloat CGPointToPixel(CGFloat point) {
+    return point * [UIScreen mainScreen].scale; // FIXME: multiple screens?
+}
+
+- (bool)isPointOnVMDisplay:(CGPoint)pos {
```
saagarjha: `CGRectContainsPoint`?

osy86:
> > Right, that's why I also changed the contrast so the buttons are white (which frankly I like more). I think I might merge that and then rework this so it just handles the cursor in the VM. What do you think?
> 
> Sure! I can do that if you want, it's just a matter of rolling back some of the changes I made. Should I keep the effect for the vm list? It magnifies the cell slightly on hover.

I think it's a bit confusing for the user because it lifts the entire cell but the whole cell is not clickable (plus the Edit button is not differentiated so it's not clear it's clickable). In my version, the run button and the edit button are highlighted. It also is more consistent with the pause screen where the resume button is highlighted.

osy86:
Also, I merged #192 so you can rebase from it.

Views/VMDisplayMetalViewController+Pointer.m
conath: That works, thank you.

osy86:
Also, I guess because of how the release pipeline is set up, we won't be able to make a release until https://github.com/actions/virtual-environments/issues/620 is resolved.

Views/VMDisplayMetalViewController+Pointer.m
conath: That didn't work because the rects that this function builds aren't sized properly. I didn't dive too deep into it.

Views/VMDisplayMetalViewController+Pointer.m
conath: It's a hack, and has been removed!

Views/VMDisplayMetalViewController.m
conath: It's part of UIKit, I don't know if one can weak link only parts of it.

conath:
> Also, I merged #192 so you can rebase from it.

The rebase worked as far as I can tell, nice button hover effects everywhere! And cursor still working in VM.

osy86:
Sorry, can you `git rebase master` and `git push -f` so it's easier to see the changes?

Also, I see that sometimes if you move the cursor too fast, you'll hit an edge or the toolbar and the cursor would stop. Is there any way to "capture" the cursor (in the SPICE GTK client, the cursor is forced to (0,0) every time to prevent it from leaving the window)?

conath:
> Sorry, can you `git rebase master` and `git push -f` so it's easier to see the changes?
I will clean up the history tomorrow.
> Also, I see that sometimes if you move the cursor too fast, you'll hit an edge or the toolbar and the cursor would stop. Is there any way to "capture" the cursor (in the SPICE GTK client, the cursor is forced to (0,0) every time to prevent it from leaving the window)?
The cursor is controlled by the system and can’t be captured like on desktop OS meaning I can’t set the system cursor position, only read. 🤷‍♂ And iOS stops sending cursor updates once it reaches the edge of the screen. A possible solution to the toolbar issue would be to overlay a view on top of everything that only handles cursor inputs. The way it works now is similar to the continuity cursor mode in many desktop VMs where you can seamlessly mouse away from the VM.

Another approach could be to use [UIApplicationSupportsIndirectInputEvents](https://developer.apple.com/documentation/bundleresources/information_property_list/uiapplicationsupportsindirectinputevents?changes=latest_minor&language=objc) but it seems to only help distinguish gestures, not remove the pointer limitations.

osy86:
I see, I think I have another idea which I'll work on in another branch. Basically for #122 I'm implementing qemu's `usb-tablet` option which emulates a USB tablet device. This allows SPICE to send an absolute position instead of a motion vector. I'll change it to allow the user to choose between mouse and tablet for touch, pencil, and touchpad.

conath:
@osy86 Here you are, it's much easier to see the changes now. 😆 [single commit](https://github.com/conath/UTM/commit/f0ebb25b186c71a8c9c1210cc00f4d7525bcd4a4)

Views/VMDisplayMetalViewController.m
saagarjha: I think you'd add something like `-Wl,-U,_OBJC_CLASS_$_UIPointerInteraction` to the build flags.

itselijahciali:
What if we make guest additions for OSes like Windows, capture the cursors, and carry that data to the VM? Wouldn't we be able to put it in a specific address in memory and have the host read it? From what I can see from [https://developer.apple.com/documentation/uikit/pointer_interactions](this guide,) it seems like we can make the guest OS use the Windows pointer to reduce lag, and report the cursor location to the guest server?

osy86:
That’s already what we’re doing? Unless I’m misunderstanding?

itselijahciali:
Looks like I misunderstood the thread. Sorry about that

Issue #188, opened 2020-03-27
usb-kbd: extended key scancodes cause repeated keys


osy86:
This mainly affects aarch64 virt machines which uses `-device usb-kbd`. When you press the arrow key, it sends two scancodes. For some reason, this causes the same key to be repeated over and over again.

zwerdna:
Hi! Are there any workarounds?

osy86:
Not that I know except don’t use arrow keys

PR #247, opened 2020-04-10
Fix aarch64 arrow keys

lshzh-ww:


lshzh-ww:
This fix issue #188

osy86:
These are scancodes for the numpad arrow keys. I'm not sure if they will be good for other configurations.

lshzh-ww:
> These are scancodes for the numpad arrow keys. I'm not sure if they will be good for other configurations.

A few configurations I have tested:
i386 & x86_64 with Windows xp/7
ppc64 with Ubuntu server
aarch64 with debian 

So far they all work well.

PR #262, opened 2020-04-13
gamepad: move the cursor continuously while holding the right stick.

js-john:
Thank you very much, I learned a lot from your refactored code!
But there is a problem: if timer is removed, the cursor will only move when the x or y values of the right stick changed. This means when you hold the right stick, the cursor will not move.
I want to add the timer back. Is that OK?


osy86:
I tend to try to avoid any timers as recommended by Apple: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/MinimizeTimerUse.html

Since `VMCursor` is a `UIDynamicItem` I think the best way is to use `startMovement` when thumbstick moves. Then use a `UIDynamicAnimator` to set a velocity (with 0 friction) based on how far the thumbstick is from the center. I can attempt this later if I have time or you can try it.

For scrolling, we can make a new `UIDynamicItem` to track it and share the same logic.

js-john:
> I tend to try to avoid any timers as recommended by Apple: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/MinimizeTimerUse.html
> 
> Since `VMCursor` is a `UIDynamicItem` I think the best way is to use `startMovement` when thumbstick moves. Then use a `UIDynamicAnimator` to set a velocity (with 0 friction) based on how far the thumbstick is from the center. I can attempt this later if I have time or you can try it.
> 
> For scrolling, we can make a new `UIDynamicItem` to track it and share the same logic.

ok, I am working on this.

PR #269, opened 2020-04-14
iPadOS 13.4 key press handler

### LINKED ISSUE #228: Improve keyboard event handling in iPadOS 13.4
EDIT: I accidentally added the bug label. I am not sure how it happened and I am not sure how to remove it.

**Describe the bug**
When using the iPad Pro Smart Cover Keyboard (or even the iOS screen virtual keyboard), some games on MSDOS 6.22 don’t recognize the keyboard keys in certain situations. For example, MicroProse’s F1GP (1991), the keys work naturally within the menu screens, but when I go to race, the game simply does not recognize the keys to control the car (A, Z, < and >). I can’t play DOOM either, the character does not move because the game also does not recognize the controls (but within the menu screens they do). It’s a strange behavior and I believe it has something to do with the low level instructions or something like that?

Also, in certain situations, while I am typing, one key or another gets stuck forever and the VM crashes. 

Another strange situation: If I do multitasking (example: UTM and Safari). While UTM is on background (not seen) and I am on Safari typing or clicking, if I go back to UTM it repeats by itself the same moves (at least on my VM, running MSDOS 6.22 & Windows 3.11, it did while I was on Windows). I cannot stop it, I have to force close it and hard reboot the iPad.

Mouse issues? I use the MSDOS MOUSE.COM to provide a PS/2 mouse for DOS level games, it works for all games except Carmen Sandiego. Windows has its own. The only issue is that after attempting to run Carmen Sandiego, the game is playable, but without mouse. And after playing it, the mouse stops working for all games and also Windows. I have to reboot to have it working again. I changed the mouse.com version various times, I have even tried CuteMouse (a lightweight version), but the same issues reappear. It does not happen on DOSBox and iDOS2/DOSPad for iOS. While on my VMWare VM, just running the game crashes the VM and forces reboot.

And on an unrelated note, just for comparison, I do not get these problems while using iDOS2/DOSPad on iPad Pro. Except, the mouse issue 

But for other several (good) reasons, I’d like to use UTM instead of iDOS2. These issues made me think they have something to do with the implementation of keyboard and mouse for UTM. I’d suggest to look on iDOS2/DOSPad (litchie’s github) for how its implemented. I know its very old code (iOS 10), but it still works as of today and as intended.

**To Reproduce**
Just run a VM with MSDOS 6.22 and Windows 3.11. Games tested: Where In The World Is Carmen Sandiego (1991 version), DOOM, MicroProse F1 Grand Prix, and MicroProse Grand Prix 2.
**Debug log**
[debug.log](https://github.com/utmapp/UTM/files/4435791/debug.log)

**Upload VM**
Well, my VM has a 2GB hard disk so I do not have any viable place to upload it.

osy86:
Resolves #228

This improves external keyboard on iPadOS 13.4. Key down and up events are handled directly from the OS which allows keys to be held down. Also fixed some other key-code handling issues including a proper fix to the arrow key repeat issue and the Clover/Windows key not registering.


mikakonn:
How do i install this app

PR #302, opened 2020-04-20
input: customize the minimum duration of long press gesture.

js-john:
The default 0.5 seconds may be too long for some people.


js-john:
Secrets are not passed to workflows that are triggered by a pull request from a fork.

osy86:
To get actions to pass you need to make a new github secret `XCODE_PATH` and set it to `/Applications/Xcode_11.4.app`. This is needed until Github updates their macOS environment to use 11.4 by default.

No new features are being added to v1.0. This will wait until v1.1.

PR #307, opened 2020-04-23
Sharing

sboddeus:
This might be total trash, but if you think it might be useful here is an implementation of adding sharing support to 'easily' share VMs around.


osy86:
Thanks, I’ll look at it after v1.0

PR #348, opened 2020-05-26
UI: added a button to refresh the list of VMs. 

js-john:
added a button to refresh the list of VMs.
solve the situation that UTM must be reopened after importing VMs.

IMG_CA12148A68F6-1


saagarjha:
How about pull-to-refresh?

yqs-github:
I think it can be designed as a pull-down refresh

osy86:
Yes to pull to refresh but not gonna add any new features to v1.0. But I’ll probably fix it so it’ll auto-refresh when you’re importing.

PR #389, opened 2020-06-22
Use latest libgpg-error

Starwarsfan2099:
`libgpg-error` 1.36 is not compatible with the latest version of Gawk 5 and breaks dependency compilation. Using `libgpg-error` 1.38 fixes compatibility with Gawk 5 and compiles fine.


hdjdvd:
Ygggg

PR #429, opened 2020-08-20
Update README.zh-Hans.md

yqs-github:
fix screen.png can not load in China .


wibus-wee:
I don't think it's effective. I tried, no matter how to call outside the chain, the final image links will present is a "camo.githubusercontent.com/xxxxx" format

PR #430, opened 2020-08-20
add --disable-tcg-interpreter flag to disable JIT

js-john:
Just want to get sysroot from Actions.. and check if it's a way to let UTM running on iOS 14


js-john:
Failed..

PR #444, opened 2020-09-08
UTM 2.0

### LINKED ISSUE #19: [Home] Reordering VM list

### LINKED ISSUE #67: [qemu] Out of memory faster with bulletproof JIT
From @saagarjha https://saagarjha.com/blog/2020/02/23/jailed-just-in-time-compilation-on-ios/
> Unlike processes with dynamic-codesigning which get access to [“jumbo VA spaces”](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/kern/kern_exec.c#L3061), iOS applications can normally only allocate a limited amount of virtual memory (which is determined using [a fairly elaborate calculation](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/osfmk/arm/pmap.c#L9868) based on the size of physical memory).

TCG allocates from 1024KiB to up to 2GiB of JIT cache. By default it uses ram size / 4. Because of the mirror mapping, iOS thinks we use 2x the amount of space we actually do for this JIT cache and therefore kills the app much quicker than before.

We should experiment with lowering the size of the JIT cache by default and/or allow the user to configure it. Make sure that the performance lost is less than the gains from the bulletproof JIT implementation. Otherwise, we should allow the old JIT as an option for those who value larger memory over faster speeds.

### LINKED ISSUE #109: Add/remove disks and drives when VM is running

### LINKED ISSUE #178: Race in [UTMQemuManager connect] and QAPI calls
Before the connection is successful, it is possible for a QAPI action to be called (and fail). Currently, this will trigger the error handler, but ideally we make it more user friendly by disabling UI options until QMP is connected.

### LINKED ISSUE #210: Prevent creating something with an invalid name
We should block creating images with an invalid name like "abc/def.img".

Also happens on creating VMs.

### LINKED ISSUE #255: macOS application
This, along with #132 would allow the user to run the same VM across all their Apple devices. We will likely use Catalyst and Swift UI to make development easier.
- [x] Implement VMs from any storage source (#132)
- [x] Refactor Storyboard to reuse common UI elements (#213)
- [x] Create UI for configuration on macOS
- [x] Create UI for listing VM and running VM on macOS
- [x] Enable KVM for x86_64 for near native speeds

### LINKED ISSUE #336: Template configurations
Support a new file format that allows people to share their UTM config without having to share disk images and other personal data. Maybe even maintain a database of "common" configs (Windows XP, Android, MacOS, etc) to aid in VM creation.

### LINKED ISSUE #345: No sound for DOS 6.22 and Windows 9x VMs
I did install the required drivers for Sound Blaster 16 with DOS VM and there is no sound. There's also no sound in Windows 98SE VM, with all the devices supported (Ensoniq, SB16 and AC'97). I have drivers for all three, yet no sound is coming out after installing them. I thought it was already implemented when there was given an option to select sound device in VM settings?

Both issues can be seen after a clean install of both operating systems. I have not tried higher than Windows 9x though. Also noticed that the ReactOS VM does not have sound, too.

### LINKED ISSUE #356: Sort drives list
Since drive order matters allow the user to reorganize the entries.

### LINKED ISSUE #381: [Feature Request] Even more advanced mode than the regular advanced mode
(For 2.0 maybe?)
So my idea for this is that UTM only passes the SPICE info to qemu, and you have to configure the rest, like `-cpu cortex-a57`, `.m 1G`, `-smp 8`, `-M virt` in a GUI similar to the one used for the existing additional args section. The difference here is that you have to give *all* of the args (except for SPICE stuff) yourself.

### LINKED ISSUE #408: UTM 2.0 Design Concept
**Update 9/8/20: Check out [this post](https://github.com/utmapp/UTM/issues/408#issuecomment-689132408) for the latest design.**

For UTM 2.0, we will be completely re-implementing the GUI in SwiftUI. The only exception will be the actual VM view because we use a lot of features that cannot be replicated in SwiftUI easily (gesture handlers, custom toolbar, custom keyboard, etc). For the rest of the UI, code will be shared across all platforms.

In addition to the UI overhaul, there are other design elements that needs to be changed:

- Creating a disk drive requires too many steps and is more complicated than it needs to be. We also don't provide a good way to change boot order. Also we wish to support removable disks.
- Settings that would commonly change (shared directory, removable drives, etc) should be exposed front and center without requiring the user to dig deep into settings.
- Less used/advanced settings should be hidden by default
- The Systems page is too filed, we'll refactor it into System and QEMU settings

This is a work in progress and feedback is appreciated.

(Thanks to #334 for some ideas)

## iPhone
utm-iphone-1
utm-iphone-2
utm-iphone-3

## iPad
utm-ipad-1
utm-ipad-2

## macOS
utm-macos-1
utm-macos-2

osy86:
This is a massive redesign and rewrite of the frontend with SwiftUI 2.0 which allows us to add macOS support as well. Some other features are listed below.

* Updated QEMU from 4.2.0 to 5.1.0
* Support for macOS Big Sur (closes #255 )
* Add/remove removable media while VM is running (closes #109 )
* Sorting of drives for a VM (including order the drives are attached) (closes #356 )
* Support advanced usage where only custom QEMU arguments and a few other required arguments are passed to QEMU (closes #381 )
* .utm files supports template-like configurations (closes #336 )
* Brand new design for iOS 14 and macOS 11 (closes #408 )
* VM list can be reordered (closes #19 )
* Check filename before creating VM (closes #210 )
* (iOS 14 only) Fix 2x JIT memory usage due to ptrace() hack. Since iOS 14 fixed the ptrace() hack and added JIT system functions (for correct entitlement holders), this should not be an issue for jailbroken users on iOS 14 only. (closes #67 )
* Fixed some rare crashes in the QEMU manager class (closes #178 )
* Updated QEMU should fix an issue with audio devices on older OS (closes #345 )


Configuration/UTMConfiguration.m
```diff
@@ -119,11 +88,69 @@ - (NSURL*)terminalInputOutputURL {
     return ioFile;
 }
 
+- (void)resetDefaults {
+    [self propertyWillChange];
+    _rootDict = [[NSMutableDictionary alloc] initWithCapacity:8];
```
saagarjha, line 93: Have you considered

```objc
_rootDict = [@{
	kUTMConfigSystemKey: [NSMutableDictionary new],
	// …
} mutableCopy];
```

Platform/UTMJailbreak.c
```diff
@@ -32,33 +36,62 @@ kern_return_t catch_exception_raise(mach_port_t exception_port,
...
     int mib[] = {CTL_KERN, KERN_PROC, KERN_PROC_PID, getpid()};
     struct kinfo_proc info = {};
     size_t size = sizeof(info);
     return !sysctl(mib, sizeof(mib) / sizeof(*mib), &info, &size, NULL, 0) && !!(info.kp_proc.p_flag & P_TRACED);
 }
 
-int main(int argc, char * argv[]) {
-    BOOL debugged = am_i_being_debugged();
+
+bool jb_has_jit_entitlement(void) {
```
saagarjha, line 58: Technically you should probably read your entitlements ([how iSH does it](https://github.com/ish-app/ish/blob/ce0fd919f6ca7980930c86d24e1dc480a0276e59/app/AppGroup.m#L31)) in case Apple removes the ability to unmap your JIT region and get a new one, but I guess this works for now.

Platform/UTMJailbreak.c
```diff
@@ -32,33 +36,62 @@ kern_return_t catch_exception_raise(mach_port_t exception_port,
...
+bool jb_has_ptrace_hack(void) {
+#if defined(NO_PTRACE_HACK)
+    return false;
+#else
+    int res = ptrace(-1, -1, NULL, 0);
+    if (res < 0 && errno == EINVAL) {
+        return true;
+    } else {
+        return false;
+    }
```
saagarjha, line 77: `return ptrace(-1, -1, NULL, 0) < 0 && errno == EINVAL`?

Platform/UTMJailbreak.c
```diff
@@ -32,33 +36,62 @@ kern_return_t catch_exception_raise(mach_port_t exception_port,
...
 
-void *exception_handler(void *argument) {
+static void *exception_handler(void *argument) {
     mach_port_t port = *(mach_port_t *)argument;
     mach_msg_server(exc_server, 2048, port, 0);
     return NULL;
 }
 
-static BOOL am_i_being_debugged() {
+static bool am_i_being_debugged() {
```
saagarjha, line 50: It's probably better to [check `CS_DEBUGGED` using `csops`](https://github.com/hrydgard/ppsspp/blob/cf43f2d57f5ae850572374730f1c9fb38daba01d/ios/main.mm#L29), to be honest.

saagarjha (commented):
Actual suggestions

Platform/UTMLoggingSwift.swift
```diff
@@ -0,0 +1,68 @@
...
+    private func timestamp() -> String {
+        var buffer = [Int8](repeating: 0, count: 255)
+        var timestamp = time(nil)
+        let localTime = localtime(&timestamp)
+        strftime(&buffer, buffer.count, "%Y-%m-%dT%H:%M:%S%z", localTime)
+        return buffer.withUnsafeBufferPointer {
+            $0.withMemoryRebound(to: CChar.self) {
+                String(cString: $0.baseAddress!)
+            }
+        }
```
saagarjha, line 66: Also consider

```swift
var buffer = [CChar](repeating: 0, count: 255)
var timestamp = time(nil)
let localTime = localtime(&timestamp)
strftime(&buffer, buffer.count, "%Y-%m-%dT%H:%M:%S%z", localTime)
let string = strdup(&buffer)!
defer { free(string) }
return String(cString: string)
```

MariahJauregui:
How do I download on iOS?

Platform/UTMLoggingSwift.swift
osy86: This is copied directly from https://github.com/apple/swift-log/blob/main/Sources/Logging/Logging.swift#L723

Platform/UTMJailbreak.c
osy86: I don’t like having side effects from within if-statements.

Platform/UTMLoggingSwift.swift
saagarjha: Fair enough.

Platform/UTMJailbreak.c
saagarjha: I was suggesting removing the if statement entirely, since you're returning the condition you're checking for.

Platform/UTMJailbreak.c
osy86: Yeah, we'll cross that bridge when we get there. I don't want to add that much code.

Platform/UTMJailbreak.c
```diff
@@ -41,17 +45,26 @@ kern_return_t catch_exception_raise(mach_port_t exception_port,
     return KERN_FAILURE;
 }
 
+static void init_sys_functions(void) {
+    static bool init_done = false;
+    if (!init_done) {
+        csops = dlsym(dlopen(NULL, RTLD_LAZY), "csops");
+        exc_server = dlsym(dlopen(NULL, RTLD_LAZY), "exc_server");
+        ptrace = dlsym(dlopen(NULL, RTLD_LAZY), "ptrace");
```
saagarjha, line 53: Have you considered weak-linking these and letting the dynamic linker do this work for you?

osy86: Well I got the code from you: https://github.com/hrydgard/ppsspp/commit/99c631c79113969411c6ce24a4ab80f1d89b7369

Why even weak link? Previously I just had them linked and it works fine. I thought you had some more subtle reason.

saagarjha: Huh, so you did. I seem to recall using `dlsym` so I could name my function pointers the same as library functions, but I don't know why I didn't just declare a prototype for them… If you can just link directly, I would say go for it and not to even bother with weak linking. If Apple removes those symbols from the list of things they'll allow you to link against, we can dive more into making weak linking work.

PR #464, opened 2020-09-20
Psychic Paper implementation + fix JIT

### LINKED ISSUE #459: QEMU exited with an error:Could not allocate dynamic translator buffer
QEMU exited with an error:Could not allocate dynamic translator buffer
utm version:2.0b2

osy86:
Implement APRR support for < iOS 14, Psychic Paper for < iOS 13.5, and fix JIT errors (fixes #459)


Platform/UTMJailbreak.m
```diff
@@ -56,15 +82,99 @@ static bool am_i_being_debugged() {
...
+}
 
 bool jb_has_jit_entitlement(void) {
-    void *addr = mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_ANON | MAP_PRIVATE | MAP_JIT, -1, 0);
-    if (addr != MAP_FAILED) {
-        munmap(addr, PAGE_SIZE);
-        return true;
-    } else {
-        return false;
+    static NSDictionary *entitlements = nil;
```
saagarjha, line 173: static variables are zero-initialized by default, FWIW.

patches/qemu-5.1.0.patch
```diff
@@ -3226,6 +3311,183 @@ index 09faa69a54..44d1f3644d 100755
...
++static __attribute__ ((__always_inline__)) bool jit_write_protect_supported(void)
++{
++    return false;
++}
++
++static __attribute__ ((__always_inline__)) void jit_write_protect(int enabled)
++{
++}
++
++#endif
```
saagarjha, line 3406: Consider using `os_thread_self_restrict_rwx_is_supported`/`os_thread_self_restrict_rwx_to_rx`, they backdeploy to earlier iOS versions.

saagarjha (commented):
A bit slow, but you may still find this useful.

Platform/UTMJailbreak.m
osy86: I know but I want to be explicit here. Sometimes readability is more important than succinctness.

However, this does mean the code I copy pasted at line 108 should be changed as well.

patches/qemu-5.1.0.patch
osy86: Where is that defined?

Platform/UTMJailbreak.m
```diff
@@ -56,15 +82,99 @@ static bool am_i_being_debugged() {
...
+    // Inspired by codesign.c in Darwin sources for Security.framework
+    
+    // Find our mach-o header
+    Dl_info dl_info;
+    if (dladdr(app_entitlements, &dl_info) == 0)
+        return nil;
+    if (dl_info.dli_fbase == NULL)
+        return nil;
+    char *base = dl_info.dli_fbase;
+    struct mach_header_64 *header = dl_info.dli_fbase;
```
saagarjha, line 121: Consider

```
#include <mach-o/ldsyms.h>

struct mach_header_64 *header = &_mh_execute_header;
```

patches/qemu-5.1.0.patch
saagarjha: Internal SDK, you'll have to forward declare it.

osy86: Just tried and it didn't link on iOS 13.4.1

Platform/UTMJailbreak.m
osy86: I got the code from https://github.com/ish-app/ish/blob/master/app/AppGroup.m I don't want to make too much changes, since I want to easily copy-paste changes from iSH if needed in the future. Also, it seems like we need `dl_info` anyways for `base`, so I don't think we would gain much here. And since this extern variable starts with an underscore, I'm afraid that it might get changed or removed in the future without warning.

saagarjha: FWIW:

Screen Shot 2020-09-20 at 15 45 03

Is `base` just not the address of the thing? (Also, there's `_NSGetMachExecuteHeader` from <crt_externs.h>, if that makes you more confident in this continuing to exist.)

patches/qemu-5.1.0.patch
osy86: `grep -r 'os_thread_self_restrict_rwx_to_rx' /Applications/Xcode.app` on Xcode 12 returned no results.

saagarjha: Actually, it's probably a macro. Nevermind then.

Platform/UTMJailbreak.m
osy86: I’ll look into that.

Issue #220, opened 2020-04-03
Support external monitor


osy:
iPad allows an external monitor to be used as a separate display.

conath:
I have recently implemented external screen support in my app, Present Anything. I can make it happen for UTM app, but I have a few questions. The iOS external screen API (including on iOS 14) is UIKit based, you basically provide a UIWindow to the external screen to display. It is possible to change the resolution of the ext. screen (iOS defaults to the max supported by the device and screen).

1. Which View or ViewController renders the VM screen?  
2. Is it possible to remove this view from the window and move it to the external UIWindow? (might need to find that out myself)  
3. How is the touchscreen feature implemented? Like, is it the same view that does the VM rendering or another one? (that would be better)

osy86:
I see a couple of main steps mostly from reading https://www.spice-space.org/multiple-monitors.html which gives a high level view of how SPICE handles multiple monitors.

CSDisplayMetal is the interface between libspice-gtk and the Metal rendering. So here, we have to handle notifying SPICE server that a new display arrived with some dimension. This is also where the pixel framebuffer is copied to a Metal texture. That would have to be extended to handle multiple textures representing each display and/or a mapping of texture to displays if we support display mirroring. We also need some sort of syncing with CSInput which handles the cursor position.

UTMRenderer does the Metal rendering which renders both the display and cursor through delegate calls to CSDisplayMetal and CSInput. A MTKView uses UTMRenderer. VMDisplayMetalController hosts the MTKView. VMDisplayMetalController also does the touch inputs, keyboard handling, etc. So VMDisplayMetalController (or VMDisplayTerminalController) would be good candidates to move to UIWindow.

Without SPICE agent installed, the only thing we can do is mirror or move the display to the external display. For extending the display, the guest has to be aware of the SPICE protocol. I think in the case of tablet/pencil mode cursor, it is fine to not be able to input on the external display.

Also, for new files and features, I would prefer moving to Swift from this point on.

conath:
I have added experimental iOS external screen support, with new files written in Swift. I need help with the external VMDisplayMetalViewController to get mirroring and (potentially) multiple displays working. [See my pull request](https://github.com/utmapp/UTM/pull/484).

Additionally, currently the aspect ratio of the VM display becomes distorted when it moves back from external to internal screen.

Here'a a [very short video](https://youtu.be/UgyoEKfqZiw) demoing Windows XP on an external screen connected to my iPad Pro. 😄

PR #484, opened 2020-10-09
Add iOS External Screen support

conath:
This is currently very WIP! This is eventually intended to implement #220. 

Current status: if an external screen is connected, the VM display will be moved there. This breaks touch and cursor input to the VM, which will require separating the touch handling from the `MTKView to fix`. As a 'band aid' fix, I have redirected touch handling to the VM container view (`VMDisplayMetalViewController+Touch.m`), which enables a small portion of the screen to be used to control the VM pointer.

Help needed with mirroring mode: I have attempted to implement a display mirroring mode by introducing a second `MTKView` on a new `VMExternalDisplayMetalViewController.swift` and setting its `UTMRenderer`, `sourceDisplay` and `sourceInput` the same as the internal ones. The relevant code is in `ExternalScreenController.swift` line 126. Unfortunately with regards to the `UTMRenderer` and `MTKView`, I have no idea what I'm doing. So if you use this mode by uncommenting it, that currently results in a blank screen.

### DIFF
--- a/CocoaSpice/CSDisplayMetal.h
+++ b/CocoaSpice/CSDisplayMetal.h
@@ -37,6 +37,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (id)initWithSession:(nonnull SpiceSession *)session channelID:(NSInteger)channelID;
 - (void)updateVisibleAreaWithRect:(CGRect)rect;
 - (void)requestResolution:(CGRect)bounds;
+- (void)requestResolution:(CGRect)bounds monitorID:(NSInteger)monitorID;
 
 @end
 
--- a/CocoaSpice/CSDisplayMetal.m
+++ b/CocoaSpice/CSDisplayMetal.m
@@ -408,13 +408,18 @@ static void cs_channel_destroy(SpiceSession *s, SpiceChannel *channel, gpointer 
 }
 
 - (void)requestResolution:(CGRect)bounds {
+    [self requestResolution:bounds monitorID:self.monitorID];
+}
+
+- (void)requestResolution:(CGRect)bounds monitorID:(NSInteger)monitorID {
+    
     if (!_main) {
         UTMLog(@"ignoring change resolution because main channel not found");
         return;
     }
-    spice_main_channel_update_display_enabled(_main, (int)self.monitorID, TRUE, FALSE);
+    spice_main_channel_update_display_enabled(_main, (int)monitorID, TRUE, FALSE);
     spice_main_channel_update_display(_main,
-                                      (int)self.monitorID,
+                                      (int)monitorID,
                                       bounds.origin.x,
                                       bounds.origin.y,
                                       bounds.size.width,
--- a/Managers/UTMSpiceIO.h
+++ b/Managers/UTMSpiceIO.h
@@ -35,6 +35,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)init NS_UNAVAILABLE;
 - (instancetype)initWithConfiguration: (UTMConfiguration*) configuration port:(NSInteger)port NS_DESIGNATED_INITIALIZER;
 - (void)changeSharedDirectory:(NSURL *)url;
+- (CSConnection*)getSpiceConnection;
 
 @end
 
--- a/Managers/UTMSpiceIO.m
+++ b/Managers/UTMSpiceIO.m
@@ -182,6 +182,10 @@ typedef void (^connectionCallback_t)(BOOL success, NSString * _Nullable msg);
     }
 }
 
+- (CSConnection *)getSpiceConnection {
+    return _spiceConnection;
+}
+
 #pragma mark - CSConnectionDelegate
 
 - (void)spiceConnected:(CSConnection *)connection {
@@ -216,6 +220,9 @@ typedef void (^connectionCallback_t)(BOOL success, NSString * _Nullable msg);
             self.connectionCallback(YES, nil);
             self.connectionCallback = nil;
         }
+    } else {
+        // TODO external
+        [[NSNotificationCenter defaultCenter] postNotificationName:@"externalDisplayAdded" object:nil userInfo:@{@"display":display, @"input":input}];
     }
 }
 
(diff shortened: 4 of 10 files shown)


osy86:
Yeah, I kinda regret this code but the way vmDisplay and vmInput are set is:

1. `cs_display_monitors` calls the delegate method `spiceDisplayCreated` with `CSDisplayMetal` and `CSInput` created for the new display
2. In UTMSpiceIO, `spiceDisplayCreated` sets properties on its delegate `_delegate.vmDisplay` and `_delegate.vmInput`. In order to prevent race conditions, UTMSpiceIO's `setDelegate` also sets those two properties.
3. UTMSpiceIO's delegate is VMDisplayMetalViewController. In `transitionToState` (called by UTMVirtualMachine) for `kVMStarted`, the two `UTMRenderSource` are passed to the renderer. In order to prevent race conditions (the `kVMStarted` message is dropped), this is also done in `viewDidLoad`.

So when you copy `self.vmDisplay` and `self.vmInput` in `viewDidLoad`, you are likely just copying nil at this point. You either have to re-pass them at `kVMStarted` or set up a KVO on those two.

If I have time, I might go through and re-arch this who procedure to be more understandable. This bad design was a result of incrementally adding code as new features were introduced.

conath:
You are right, the values are still nil when the view loads. I changed the implementation to grab the `UTMRenderSource` references after the VM starts, which worked. I then tried to create a second MTKView on the external view controller with the same render source as internally. However as soon as I set the delegate in the next draw call, it crashes. I now assume that a `UTMRenderSource` can't be used with two `MTKView`s. So I abandoned that idea and focused on creating an extended display mode.

I read through some of the CoreSpice code and have a gist of what happens to create the virtual display. However I don't see anywhere where I could call a function to create another display. When a new display channel is needed that seems to be signaled by Spice on `"channel-new"` with type display. How can I cause that signal? Am I understanding it correctly in the first place?

osy86:
Not 100% sure on this but look at `requestResolution` in CSDisplayMetal. It's currently setting for `self.monitorID` which afaik is always 0 right now. But it seems like you can pass in 1,2,...,15 to configure up to 16 displays. `spice_main_channel_send_monitor_config` will issue the `VD_AGENT_MONITORS_CONFIG` request. If the guest agent responds with `MONITORS_CONFIG` this is seen by `cs_display_monitors` in CSConnection. There is already code here to setting up a new `CSDisplayMetal` and `CSInput` but I don't think it's ever been tested. Once you have those you can create a new UTMRenderer and so on.

osy86:
That happens normally. libspice-gtk should automatically send the monitor config once the agent does connect (SPICE guest tools needs to be installed). Calling `spice_main_channel_send_monitor_config` is for the race condition where the agent is already connected before we called `spice_main_channel_update_display_enabled`.

conath:
Oh alright. I'll try it again with the prebuilt VM that includes the spice tools.

conath:
Just tried it with the "Debian 10.4 (LDXE) ARM" vm from the website, which includes SPICE tools. The output after the `requestResolution` call is now:

```log
qemu-system: info: GSpice: channel-main.c:1131 main-1:0: sending new monitors config to guest
qemu-system: info: GSpice: channel-main.c:1148 main-1:0: monitor #0: 1366x1024+0+0 @ 32 bpp
qemu-system: info: GSpice: channel-main.c:1148 main-1:0: monitor #1: 1920x1080+0+0 @ 32 bpp
qemu-system: info: GSpice: channel-main.c:1066 #0 +0+0-1366x1024
qemu-system: info: GSpice: channel-main.c:1066 #1 +1366+0-1920x1080
qemu-system: info: Spice: reds.c:1158:reds_on_main_agent_monitors_config: monitors_config->num_of_monitors: 16
```

However `cs_display_monitors` is not called.

osy86:
Do you get the log message `received new monitors config from guest`?

conath:
No, I don't get that message.  I tried to run `xrandr -q` and `grander -d :1` which both gave me `unable to open display`.

osy86:
> No, I don't get that message. I tried to run `xrandr -q` and `xrandr -d :1` which both gave me `unable to open display`.

`sudo` ?

conath:
CF6539FA-B4E5-4B6B-ADD8-F2A0AF3C5CD0

Same with `sudo`

osy86:
Just to be clear, auto-resolution changes do work? (i.e. rotating the screen)

conath:
No, that isn‘t working. In VM Display settings, Fit to screen is enabled, Retina mode is disabled. I‘m on iPad Pro 12.9 2nd gen, jailbroken on iOS 13.3. The VM is the "Debian 10.4 (LDXE) ARM" from the website (on the website it says that auto resolution change is not supported due to a bug, is that up to date?)

Maybe you could try this branch on one of your devices & VMs?

osy86:
Still trying to figure out why the second display won't show up but I've ran into a couple of other issues:

I've gotten rare crashes in `drawRegion`. The cause may be that `_canvasData` is assumed to be locked by `_drawLock`. However, it is valid for libspice to return the same `_canvasData` to two different monitors (they could share a canvas). Therefore either the `_drawLock` has to be shared as well or a separate lock should be created.

Also, it seems like when I rotate the iPad (simulator) or use split view, it no longer sends the right resolution to SPICE.

osy86:
Oh. Duh. We only have one VGA device.

https://people.freedesktop.org/~teuf/spice-doc/html/ch02s07.html

conath:
I tried the suggestion from the website you linked to, which was to add the GEMU argument “-vga qxl -device qxl”.

That results in the following error message when the VM starts:
1964A8E9-62FC-4944-9F23-D7FA145137B4

brunocastello:
Hi I am just curious. What's the current status on this? Is it working? If yes, how does it work, what are the current caveats, issues, supported VM OSes, requirements... ? A Samsung Dex like experience running a VM on my iPhone would be amazing IMO.

conath:
> What's the current status on this? 

It's not working and needs further development to get to a working state.

osy:
The rework in CocoaSpice on the dev branch should solve some of the issues but also require a lot of refactoring to be done.

osy:
External displays have been implemented in 4.0.0

PR #487, opened 2020-10-10
Add instructions for tether launching without mac

asdfugil:


Assfugil:
that's about it

demhademha (changes requested):
You've got a typo, it needs to be libimobiledevice, then the link

Assfugil:
Is there any problems?

conath:
Does AltStore on Windows support AltJIT?

PR #517, opened 2020-11-06
Fix a crash when removing drives with no backing

saagarjha:


osy86:
Sorry messed up the merge, but it's in.

PR #709, opened 2020-11-27
[iOS] Support app sandbox for removable drives and sharing

osy86:
This is my attempt at #531


Managers/UTMVirtualMachine+Drives.m
```diff
@@ -73,19 +73,18 @@ - (BOOL)ejectDrive:(UTMDrive *)drive force:(BOOL)force error:(NSError * _Nullabl
...
     if (!bookmark) {
         if (error) {
             *error = [NSError errorWithDomain:kUTMErrorDomain code:-1 userInfo:@{NSLocalizedDescriptionKey: NSLocalizedString(@"Failed create bookmark.", "UTMVirtualMachine+Drives")}];
         }
         return NO;
     }
-    if (![self checkSandboxAccess:url error:error]) {
-        return NO;
-    }
     [self.viewState setBookmark:bookmark path:url.path forRemovableDrive:drive.name persistent:NO];
```
conath, line 91: You're temporarily saving the bookmark with `persistent:NO` to update the view state. However it is saved across UTM restarts anyway. Maybe you could rename the flag to make the code more readable?

conath: In line 104 you are calling the `setBookmark` with `persistent:YES`, disregarding the `persistent` parameter passed, and above that in line 101 you are passing `persistent` to `securityScoped`. I don't understand what's happening here.

conath (commented):
You did it! This both fixes changing CDs while the VM is running as well as accessing files outside the sandbox. I tested it on iPadOS 14.2 with ISOs in the Downloads directory, as well as sharing the Downloads folder as shared directory for a Debian VM. 👍 

There are some minor remarks re: the `persistent` flag that we had previously discussed. I don't think it is named right, and it makes the code hard to understand. Other than that it's very nice.

osy86:
Right, I mixed up persistent and securityScoped there.

osy86:
Also, I swear I wrote this exact code and it didn't work in iOS 13. Maybe I missed something though, oh well.

BluDood:
Can you compile this into an .ipa? I don´t know how to do it myself :(

PR #712, opened 2020-11-28
Improve removable drives state and UI

conath:
The removable drives list in the VM details isn't very nice to look at currently. This PR includes some of the UI changes from my previous PR #531, and includes a fix for incorrect `UTMDrive.state`.

### DIFF
--- a/Managers/UTMVirtualMachine+Drives.m
+++ b/Managers/UTMVirtualMachine+Drives.m
@@ -88,6 +88,8 @@ extern NSString *const kUTMErrorDomain;
         return NO;
     }
     if (!self.qemu.isConnected) {
+        // On macOS, the `bookmark` will not work for the helper process.
+        // This is caught and addressed in `restoreRemovableDrivesFromBookmarksWithError`.
         [self.viewState setBookmark:bookmark path:url.path forRemovableDrive:drive.name persistent:YES];
         [self saveViewState];
         return YES; // not ready yet
@@ -129,7 +131,6 @@ extern NSString *const kUTMErrorDomain;
         BOOL persistent = NO;
         NSData *bookmark = [self.viewState bookmarkForRemovableDrive:drive.name persistent:&persistent];
         if (bookmark) {
-            NSString *path = nil;
             UTMLog(@"found bookmark for %@", drive.name);
             if (drive.status == UTMDriveStatusFixed) {
                 UTMLog(@"%@ is no longer removable, removing bookmark", drive.name);
@@ -137,8 +138,24 @@ extern NSString *const kUTMErrorDomain;
                 continue;
             }
             if (![self changeMediumForDriveInternal:drive bookmark:bookmark securityScoped:persistent error:error]) {
-                UTMLog(@"failed to change %@ image to %@", drive.name, path);
+#if TARGET_OS_IPHONE
+                UTMLog(@"failed to change %@ image", drive.name);
                 return NO;
+#else
+                // On macOS, at this point it is possible that a disk image was chosen while the VM was powered down.
+                // This results in an invalid bookmark so we need to re-run changeMediumForDrive.
+                // If the viewState contains the path for the drive, we can use that to create a new bookmark.
+                NSString* path = [self.viewState pathForRemovableDrive:drive.name];
+                if (path) {
+                    NSURL* url = [NSURL fileURLWithPath:path];
+                    if (![self changeMediumForDrive:drive url:url error:error]) {
+                        UTMLog(@"failed to change %@ image to %@", drive.name, path);
+                        return NO;
+                    }
+                } else {
+                    continue;
+                }
+#endif
             }
         }
     }
--- a/Managers/UTMVirtualMachine+Drives.m
+++ b/Managers/UTMVirtualMachine+Drives.m
@@ -45,16 +45,23 @@ extern NSString *const kUTMErrorDomain;
         drive.interface = [self.configuration driveInterfaceTypeForIndex:i];
         drive.name = [NSString stringWithFormat:@"drive%lu", i];
         NSString *path = removableDrives[drive.name];
-        if (path) {
+        if ([self.configuration driveRemovableForIndex:i]) {
+            // removable drive
             if (path.length > 0) {
                 drive.status = UTMDriveStatusInserted;
                 drive.path = path;
             } else {
-                drive.status = UTMDriveStatusEjected;
+                path = [self.configuration driveImagePathForIndex:i];
+                if (path.length > 0) {
+                    drive.status = UTMDriveStatusInserted;
+                    drive.path = path;
+                } else {
+                    drive.status = UTMDriveStatusEjected;
+                    drive.path = nil;
+                }
             }
-        } else if ([self.configuration driveRemovableForIndex:i]) { // qemu not started yet
-            drive.status = UTMDriveStatusEjected;
         } else {
+            // fixed drive
             drive.status = UTMDriveStatusFixed;
             drive.path = [self.configuration driveImagePathForIndex:i];
         }
@@ -66,6 +73,7 @@ extern NSString *const kUTMErrorDomain;
 - (BOOL)ejectDrive:(UTMDrive *)drive force:(BOOL)force error:(NSError * _Nullable __autoreleasing *)error {
     NSString *oldPath = [self.viewState pathForRemovableDrive:drive.name];
     [self.viewState removeBookmarkForRemovableDrive:drive.name];
+    [self.configuration setImagePath:@"" forIndex:drive.index];
     [self saveViewState];
     [self.system stopAccessingPath:oldPath];
     if (!self.qemu.isConnected) {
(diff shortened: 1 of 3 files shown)


Managers/UTMVirtualMachine+Drives.m
```diff
@@ -87,12 +87,16 @@ - (BOOL)changeMediumForDrive:(UTMDrive *)drive url:(NSURL *)url error:(NSError *
         }
         return NO;
     }
-    NSString *oldPath = [self.viewState pathForRemovableDrive:drive.name];
-    [self.viewState setBookmark:bookmark path:url.path forRemovableDrive:drive.name persistent:NO];
     if (!self.qemu.isConnected) {
+        [self.viewState setBookmark:bookmark path:url.path forRemovableDrive:drive.name persistent:YES];
```
osy86, line 91: This still won't work for macOS. To get a persistent bookmark, you need to get it from the helper process, which at this point isn't launched yet.

We can make it so it saves the settings only for iOS (which doesn't have this requirement) but it would be inconsistent with macOS. Or we can refactor the code so it will always call `[self.system accessDataWithBookmark]`.

Platform/Shared/VMRemovableDrivesView.swift
```diff
@@ -18,7 +18,7 @@ import SwiftUI
 
 @available(iOS 14, macOS 11, *)
 struct VMRemovableDrivesView: View {
-    let vm: UTMVirtualMachine
+    @State var vm: UTMVirtualMachine
```
osy86, line 21: Is this needed in other places where I have `let vm` as well? My understanding is that as long as the final path is an ObservableObject we should be good? At least that would explain it working in other places...

conath: You're right, this is not needed. It is a remnant from debugging.

Managers/UTMVirtualMachine+Drives.m
conath: Would it work if I change it to save just the drive path on macOS:

```objc
    if (!self.qemu.isConnected) {
#if TARGET_OS_IPHONE
        [self.viewState setBookmark:bookmark path:url.path forRemovableDrive:drive.name persistent:YES];
        [self saveViewState];
#else
        [self.configuration setImagePath:url.path forIndex:drive.index];
#endif
        return YES; // not ready yet
    } else {
```

After QEMU connects, in `restoreRemovableDrivesFromBookmarks` I could create a bookmark from the saved path.

osy: For consistency purposes I would prefer the behaviour be consistent across macOS and iOS even if it's not ideal.

Platform/Shared/VMRemovableDrivesView.swift
```diff
@@ -31,44 +31,43 @@ struct VMRemovableDrivesView: View {
...
     init(vm: UTMVirtualMachine) {
-        self.vm = vm
+        self._vm = State<UTMVirtualMachine>(initialValue: vm)
         self.config = vm.configuration
         self.sessionConfig = vm.viewState
     }
     
     var body: some View {
-        Group {
+        return Group {
```
osy, line 40: Is there a reason for the added `return` here?

Platform/Shared/VMRemovableDrivesView.swift
```diff
@@ -31,44 +31,43 @@ struct VMRemovableDrivesView: View {
...
-                    HStack {
-                        Label("Interface: \(drive.interface ?? "")", systemImage: drive.imageType == .CD ? "opticaldiscdrive" : "externaldrive")
-                        Spacer()
-                        Text(path)
-                            .lineLimit(1)
-                            .truncationMode(.head)
+            ForEach(vm.drives.filter { $0.status != .fixed }) { drive in
+                HStack {
+                    DriveLabel(drive: drive)
+                    Button(action: {
```
osy, line 57: Are you doing "Icon - Button - Path" instead of "Icon - Path - Button" for a reason? I like the latter because it's more consistent with existing UI (path - button).

Platform/Shared/VMRemovableDrivesView.swift
conath: No, there isn't. I'll remove it

Platform/Shared/VMRemovableDrivesView.swift
conath: My thinking was that I want to separate a slightly destructive action (clear) from a non-destructive action (browse). If you like it as well, I can change the shared directory UI to match. Here's how that would look (not pushed yet).
RemovableDrives - 1
RemovableDrives - 2
RemovableDrives - 3
RemovableDrives - 4

Managers/UTMVirtualMachine+Drives.m
conath: With the fix I suggested it would achieve consistent results on both platforms, even though the behind the scenes is slightly divergent. Is that not acceptable?

osy: Sure I’m fine with it but can you make sure to test everything on both iOS and macOS?

1. Load image and exit
2. Load image and start
3. Start and change image and exit
4. Start and don’t change image and exit
5. Start and clear and exit
6. Second launch, load and exit
7. Second launch, use existing and start
8. Second launch, clear and exit
9. Second launch, clear and start

And make sure it works as expected.

Platform/Shared/VMRemovableDrivesView.swift
osy: That’s a good point. But is it possible to swap the two and align the clear button? Maybe that would need a geometry reader...

conath: Sounds good, I'll see what I can come up with!

conath: I have come up with an alternative design that was easier to implement, and looks very clean. Instead of the _Browse_ and _Clear_ buttons, the "Shared Directory" label on the left is now tappable and reveals a menu with _Browse_ and _Clear_ buttons. I also changed the icons for the VM details to be the same color as the text. That makes more sense anyway because they aren't interactive.

IMG_0079

When a VM is suspended, it becomes greyed out:

IMG_0080

Tapping the button makes a menu pop up just above it, containing the _Browse_/_Clear_ buttons.

IMG_0088

If the drive is ejected or no shared path is set, the _Clear_ button isn't shown.

IMG_0087

I realize that this doesn't fulfill the goal of separating the _Clear_ from the _Browse_ button, however with the Menu approach the _Browse_ button is closer to the initial touch location than the _Clear_ button so it should be fine. I tried making the _Clear_ button's text color red but that doesn't work inside a Menu.

Managers/UTMVirtualMachine+Drives.m
conath: Unfortunately, I can't get UTM to work on my MacBook Air (Big Sur 11.1 Beta). I can't start any VMs because it can't connect to the QEMU instance (console logs many _SPICE port not in use_ messages). So I haven't made any progress on this yet.

osy:
@conath Sorry to bother you but is there an update on this? I recall you had some new design ideas but I don't think they're in this PR.

conath:
@osy Thanks for reminding me about this. I had stopped working on it because of the [macOS bookmark issue](https://github.com/utmapp/UTM/pull/712#discussion_r541934827) where the file bookmark needs to be created by the helper process. The UI work is done. I'll look into it again right now.

Managers/UTMVirtualMachine+Drives.m
conath: Sorry, I don't understand your test notes. (start/launch?)

> 1. Load image and exit
> 2. Load image and start
> 3. Start and change image and exit
> 4. Start and don’t change image and exit
> 5. Start and clear and exit
> 6. Second launch, load and exit
> 7. Second launch, use existing and start
> 8. Second launch, clear and exit
> 9. Second launch, clear and start

Here's all the test scenarios that I tested on my machine. I begin with a x86 VM created that has only a removable CD drive with no image inserted (no other boot device). Each following test continues with the `.utm` state after the previous test. Observations are in (brackets).

- start UTM, Browse for an image, exit UTM, start UTM, Start VM (CD loads correctly), exit UTM
- start UTM (image inserted), clear image, exit UTM, start UTM (no image), exit UTM
- start UTM, browse for a different image, exit UTM, start UTM, browse for first image, Start VM (CD loads correctly), exit UTM
- start UTM, start VM (CD loads correctly), eject CD, exit UTM, start UTM (no image), exit UTM
- start UTM, start VM (no bootable medium), browse for image, reset VM (CD loads correctly), exit UTM

All of the above worked as expected (macOS 11.1). The new code that makes the image persistence work is in [UTMVirtualMachine+Drives.m](https://github.com/conath/UTM/blob/8a5f2060a267fe80b0fdcd9789c747f503ee249f/Managers/UTMVirtualMachine%2BDrives.m#L145).

conath:
Here are a few screenshots that show the updated removable drives UI. The Browse/clear is now in a Menu on the left, which cleans up the look a lot. I think it looks best on iOS.

Note that on macOS, the Browse button somehow takes up the entire width despite [the Spacer() after the menu](https://github.com/conath/UTM/blob/747cd2bf9cda6a02a17923cc6f493646551e00fe/Platform/Shared/VMRemovableDrivesView.swift#L85) (it works as expected on iOS).
utm-drivesui-macos
IMG_0113
IMG_0112
IMG_0111
IMG_0110

PR #1413, opened 2020-12-20
Add SidebarCommands to be able to bring Sidebar back.

laosb:
On macOS, user is able to adjust sidebar width by dragging the separator. If the user drag it to near 0 width, it's impossible to get the sidebar back again. Adding SidebarCommands would have a command in View menu for user to toggle sidebar, as well as a key shortcut for doing so.

image


osy:
I think this may have build issues on iOS. There’s plans to introduce more menu items and this change is in the branch which I haven’t pushed yet.

PR #2349, opened 2021-02-28
Refactor confirmation alerts to fix #2337

### LINKED ISSUE #2337: Delete confirmation always shown twice
**Describe the issue**
When deleting a VM on macOS (from the right click menu), the *Do you want to delete this VM and all its data?* alert shows up again after one of the options is clicked. If the *Delete* button is clicked on the first alert, the VM is deleted. Then, if *Delete* is clicked again on the second alert this results in another alert telling you that the VM could not be deleted.

**Configuration**
* UTM Version: latest
* OS Version: latest
* Intel or Apple Silicon? Intel

conath:


conath:
It seems that adding the context menu modifier in ContentView.swift line 86 breaks dismissing the VM Config modal.

osy:
It seems like in addition to fixing the double alerts, you've also made it automatically select the last VM whenever something is deleted? Is that right? What is the reason for this? Have you tested it on iOS where I think it will navigate to the details view whenever you delete a VM?

conath:
I tried to make it go back to the welcome screen after a VM is deleted. And it’s supposed to show the new VM details when you create a VM. I can test it on iPhone later this week, it is working on Mac.

osy:
I really dislike the idea of adding lots of code to workaround a SwiftUI bug because from experience, whenever an update is released, it might break the workaround. In addition I've also had in the past, the workaround creating some new edge condition that breaks (often on a different platform). It's just a pain to deal with and I think I would rather Apple fix SwiftUI. That being said I filed FB9030225 for this bug and attached the following minimally reproducible example:

```swift
struct ContentView: View {
    @State private var alertShown = false
    var body: some View {
        NavigationView {
            List {
                NavigationLink(destination: Button(action: { alertShown.toggle() }, label: {
                    Text("Show Alert")
                })) {
                    Text("Click Here")
                }.alert(isPresented: $alertShown, content: {
                    Alert(title: Text("Alert!"))
                })
            }
        }
    }
}
```

PR #2361, opened 2021-03-03
Fix resolution scaling from Cocoa to guest

akihikodaki:


osy:
Can you explain your change? Native resolution means unscaled on the host side, so it should appear "smaller" if you have a Retina display. The purpose of this is if you are running an OS which has its own scaling (i.e. Windows with 200% rendering), it will not interfere with macOS's own scaling.

akihikodaki:
Indeed, everything looks smaller if I run Ubuntu with native resolution. This pull request does not change that.

First, let me clarify terminologies.

- "Cocoa resolution" is what Cocoa operates on.
- "Native resolution" is the resolution in pixels on the actual display.
- "Guest resolution" is the resolution in pixels which the guest recognizes.

On a Retina display, the native resolution is much larger than one of a classic display, but the physical size (in inches, for example), is consistent with a classic display.

Now let's assume `isAlwaysNativeResolution` is true. If it is true, the guest resolution is expected to equal to the native resolution.

Actually there are two ways to synchronize the guest resolution and native resolution:
1. Cocoa receives a frame from the guest. The frame size is represented as guest resolution, and it must be converted to Cocoa resolution.
2. Cocoa notifies the guest the change of the window size. The window size is represented as Cocoa resolution, and it must be converted to guest resolution.

This pull request fixes 2. Since the guest resolution must be equal to the native resolution, we scale Cocoa resolution by `currentScreenScale`.

osy:
I see thanks for the explanation.

dejavuwl:
hello everyone, does this [https://github.com/utmapp/UTM/discussions/3319](this) fixed by this commit? @akihikodaki

akihikodaki:
Maybe related, but it is a distinct problem. This was merged long before, and is included in 2.4.1.

PR #2377, opened 2021-03-13
iOS: use GCMouse API on iOS 14 and above

js-john:
With this api applied, the mouse in virtual machine performs like a real mouse
the mouse cursor will not get out of the screen of virtual machine, or trigger the notification center and multi-task window

https://user-images.githubusercontent.com/9637998/111026862-b5958c80-8427-11eb-9a1e-71ea4ec3eead.mov


milkly1235:
Use

milkly1235:
Use

PR #2392, opened 2021-03-20
Add a tip to tell user UTM for iOS is free.

### LINKED ISSUE #2391: Add a tip to tell user UTM for iOS is free.
These days, some people Sell UTM on taobao.com. I think we need to add a tip to tell user UTM for iOS is for free.

IMG_EA479E3A1360-1

ty-yqs:
Fixes #2391


saagarjha:
Not sure how useful this would be considering that it is trivial to patch out…

osy:
I tend to agree. While I appreciate the effort and concern, I think it's possible to weed out these scams and the best bet is to just ignore it.

PR #2440, opened 2021-04-10
Add support for enterprise certificates

PS5GAMERS:
I have a App Store account and i want to request Apple the UTM publication. I think that Apple will accept because "An iPad it's a computer". Can resolve the bug of the enterprise certificates? Pls!


osy:
Wrong place. Check discussions tab. And the answer is no.

PR #2451, opened 2021-04-18
Support R/W accessing image files on external drives

conath:
This is a proof of concept that changes the "Import Drive" feature to instead link to the drive file. Additionally, qemu runtime errors are ignored to prevent the VM from quitting when I/O errors occur. (this causes instability because the errors that are ignored are not filtered)

These changes allow image files on external drives to be read and written to in the VM. For example, I was able to attach a QCOW2 image to a Windows VM and copy a file to it.

### DIFF
--- a/Managers/UTMQemuSystem.m
+++ b/Managers/UTMQemuSystem.m
@@ -233,6 +233,8 @@ static size_t sysctl_read(const char *name) {
         if (hasImage) {
             if ([path characterAtIndex:0] == '/') {
                 fullPathURL = [NSURL fileURLWithPath:path isDirectory:NO];
+            } else if ([path hasPrefix:@"file://"]) {
+                fullPathURL = [NSURL URLWithString:path];
             } else {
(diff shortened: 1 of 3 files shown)


ferdyan86 (approved):

phirestalker:
This would be great. I have a large qcow2 image.

frozen-eye:
Could you please prioritize this feature as it seems very important for developers/testers who does not have lot's of free space on primary SSD because of very high price for 512GB/1T macs?
Thanks in advance!

frozen-eye:
> This would be great. I have a large qcow2 image.

my one for development requires 200GB+ qcow (SONiC for arm64 development) :(

osy:
@frozen-eye its already implemented in the lasted two beta

osy:
I don't think that's what the person above is asking for though

frozen-eye:
> @frozen-eye its already implemented in the lasted two beta

thanks! 3.1.2 beta used. working for me 👍

osy:
I think we discussed this and found that having both removable drive support, running as shortcut, and this would be rather confusing. If you want to store image files on an external drive, either use a removable drive or store the .utm on an external drive.

PR #2478, opened 2021-04-30
Improve building instructions

jkcoxson:
There are missing steps in the instructions, this just makes it clear what you have to do.


osy:
https://github.com/utmapp/UTM/tree/dev

PR #2480, opened 2021-05-02
Add the "Download Spice Guest Tools" button

ty-yqs:
Add the "Download Spice Guest Tools" button.


conath (commented):
Thanks for this pull request. Your idea about including an easy way to find the Spice Guest Tools is a nice user experience benefit, however I believe the start screen of the app is not the best place for it. (currently there is no way to return to this screen apart from relaunching the app)

In a separate pull request #2498, I have taken your idea and changed it to instead link to the user guide and FAQ / support.

ty-yqs:
That’s OK. So, should I close this pull request?

PR #2498, opened 2021-05-08
Update Welcome Screen links

conath:
Update User Guide link on iOS
Add Support link on macOS and FAQ link on iOS

and remove some unnecessary `Spacer()`s

Only thing missing is the localization.

### DIFF
--- a/Platform/Shared/VMPlaceholderView.swift
+++ b/Platform/Shared/VMPlaceholderView.swift
@@ -23,30 +23,26 @@ struct VMPlaceholderView: View {
     
     var body: some View {
         VStack {
-            Spacer()
             HStack {
-                Spacer()
                 Text("Welcome to UTM").font(.title)
-                Spacer()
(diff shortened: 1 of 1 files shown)


osy:
1. I think the macOS guide applies to iOS 14+ version as well. (at least it's better than the iOS 13 guide which is useless and is why I didn't link to it. I just never had time to write the v2 guide).
2. Similarly the support page for macOS also applies.

I think therefore it may make sense to just link to the macOS page even for iOS. Unless you read the address bar, it won't be too much of an issue. Maybe at some point I'll unify the two sites.

conath:
I have updated it to include your feedback. It looks like this now on all platforms:
Bildschirmfoto 2021-05-09 um 21 26 23

Notice that I changed the gallery icon to match the circular icon in the other buttons. 

I've cancelled the CI run for the last push because it's such a minor change.

PR #2499, opened 2021-05-08
Add workaround for SwiftUI bug in VMRemovableDrivesView

conath:
## Reproducing the bug
1. Select a VM with a removable drive
2. Tap the browse button to open the Files app modal 
3. Tap outside the modal to dismiss it
4. Tap the browse button again
Expected:
5. The modal re-appears
Actual:
5. Due to a SwiftUI bug the modal doesn't reappear.
The technical reason for the bug is that the Files importer doesn't get canceled properly when the user taps outside the modal. Instead, the presented ViewController is dismissed and the necessary `isPresented = false` callback inside SwiftUI doesn't happen.

## Workaround
We can catch this erroneous behavior by, at step 4, checking if the modal is already supposedly presented (`diskImageFileImportPresented`). This is normally impossible because the modal blocks interaction with the view below it (and thus the Browse button). That is how we know that the bug has happened earlier. Now we need to reset the `diskImageFileImportPresented` to `false` and, after the view updates, set it to `true` again to show the modal. This is achieved with an additional `@State` `Bool` just for the workaround.

## Why is this needed
This bug affects us in multiple places, including the Drives configuration and the Import UTM Virtual Machine menu. However, the bug is less annoying in those places, as those only require a single click on a button (which can easily be retried). Therefore, in those places, the current `.toggle()` behavior suffices: the `isPresented` is set to true on the next click. However with the removable drives view, I believe that the fact that the Browse button is tucked away in a menu means this bug is extra annoying.


osy:
I think it's cleaner to use `.onChange(of: workaroundFileImporterBug)` near where the bug occurs instead of at the beginning with a new return.

conath:
That's a great idea and makes more sense when you read the code. I've changed the workaround to be a `.onChange` on the `Button` where the bug is triggered.

osy:
Is `DispatchQueue.main.async` needed here? My understanding of SwiftUI is that you are always allowed to modify the state from inside a View.

PR #2563, opened 2021-05-23
Added note on Prebuilt Dependencies for iOS Development

mikegrima:
- Added a note on pre-built dependency versioning for clarity


osy:
I think this only really applied to v2.1.0 because of the massive changes in how the dylibs are imported. I don't anticipate that happening again.

mikegrima:
It's probably a good practice to do that if you are trying to make a stable release build.

PR #2640, opened 2021-06-28
Add ios_simulator-tci to the build matrix so we start getting artifacts for iOS SE Simulator 

ntindle:
Adds the ios_simulator-tci to the build matrix so we get sysroot assets to be able to fill out the full [dependencies](https://github.com/utmapp/UTM/blob/master/Documentation/iOSDevelopment.md#dependencies) table from the iOS dev docs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         arch: [arm64, x86_64]
-        platform: [ios, ios_simulator, ios-tci, macos]
+        platform: [ios, ios_simulator, ios_simulator-tci, ios-tci, macos]
         exclude:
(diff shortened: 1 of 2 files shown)


osy:
I think I didn’t build this because it uses a different interpreter (ARM64 now uses the threaded interpreter that ktemkin wrote only for ARM64) so it wasn’t worth the extra 2 hours build time.

ntindle:
Fair enough. I just ran across it missing when verifying I setup the dev env correctly for all the projects.

ntindle:
Lmk a solid yes/no so I can close this/ look forward to getting it merged.

osy:
I’m going to say no because of the limited use case and that I’m already bumping into GitHub resource limits.

PR #2670, opened 2021-07-17
Add URL scheme

conath:
A UTM scheme `utm://` is implemented that offers commands for managing VMs.

For example, the URL `utm://start?name=Ubuntu` would open UTM and start a VM called "Ubuntu".

## Commands

Proposed list of commands and their parameters, in pseudocode:

- start(name)
- stop(name)
- restart(name)
- suspend(name)
- resume(name)

Possible additional commands:
- keystroke(name, string)
- click(name, x, y)

## Status

I have implemented the start command using UTMData.run and will implement the remaining ones as well.

Not sure about the additional commands, and please feel free to comment your ideas.

### DIFF
--- a/Platform/Shared/ContentView.swift
+++ b/Platform/Shared/ContentView.swift
@@ -84,7 +84,7 @@ struct ContentView: View {
         }.overlay(data.showSettingsModal ? AnyView(EmptyView()) : AnyView(BusyOverlay()))
         .optionalWindowFrame()
         .disabled(data.busy && !data.showNewVMSheet && !data.showSettingsModal)
-        .onOpenURL(perform: importUTM)
+        .onOpenURL(perform: handleURL)
         .handlesExternalEvents(preferring: ["*"], allowing: ["*"])
         .onReceive(NSNotification.NewVirtualMachine) { _ in
             data.newVM()
@@ -145,6 +145,16 @@ struct ContentView: View {
         }
     }
     
+    private func handleURL(url: URL) {
+        if url.isFileURL {
+            importUTM(url: url)
+        } else if let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
+                  let scheme = components.scheme,
+                  scheme.lowercased() == "utm" {
+            handleUTMURL(with: components)
+        }
+    }
+    
     private func importUTM(url: URL) {
         guard url.isFileURL else {
             return // ignore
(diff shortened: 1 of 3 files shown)


conath:
There's a somewhat reproducible crash in the downloadURL feature I added. Here's the crash log, maybe it has something to do with the way I am using ProgressView. 
[combine-progressview-crash.log](https://github.com/utmapp/UTM/files/6871751/combine-progressview-crash.log)

conath:
I have worked around the crashing issue I mentioned previously by changing my implementation as explained in 
e0b9f29.

This means that in addition to the simple commands like `start`, `stop`, etc. the `downloadVM` is now working as intended.

Here is a screen recording of the  `downloadVM` feature, showing initiating and cancelling a download and also what happens when a download finishes successfully:

https://user-images.githubusercontent.com/12073163/128438199-10950a53-4fa6-482a-af2b-4b7616447600.mov

One visual glitch remains: If you cancel a pending VM and then start another download, the old (cancelled) view is shown. I don't know why this happens. The screen recording above shows this bug triggered by the steps from 0:07-0:12, with the glitched view visible from 0:12-0:28.

conath:
I decided to implement the two remaining features and rebased the changes on top of master, to make it easier for me to develop.

The `click` action is done, and the `sendText` works but only in console mode. It doesn't work properly on macOS in graphics mode (i. e. non-console VMs) because I wrote a replacement for the char-to-keycode code in Swift and that is apparently not correct. I'll take another look at it soon.

conath:
All basic and some _complex_ keystrokes are now supported by the ~~keystroke~~ sendText action. Some examples:

`utm://sendText?name=Ubuntu%20ARM%202.2.0&text=asdf` types "asdf" into the VM "Ubuntu ARM 2.2.0" if it is running.
`utm://sendText?name=Ubuntu%20ARM%202.2.0&text=%01` presses ctrl-A in the VM (see [ASCII Control Codes](https://jkorpela.fi/chars/c0.html); codes 0-31 and 127 (esc) are supported)
`utm://sendText?name=Ubuntu%20ARM%202.2.0&text=%0A` presses return/enter in the VM (see [ASCII Code Table](https://ascii-code.net); most of the ASCII codes are supported)

Non-ASCII is also supported, but depends on the keyboard layout your Mac and guest uses! If your Mac has QWERTY but your guest has AZERTY set, passing "QW" to sendText will result in "AZ"!

PR #3055, opened 2021-08-26
iOS AltKit support

JoeMatt:
## Overview

Attempts to auto-enable JIT support for users using AltServer.

- Adds AltKit SPM Package (uses my custom fork until they fix SPM support, git tags)
- Adds code to enable JIT if AltServer found
- Adds Bonjour info to Info.plist

AltKit: https://github.com/rileytestut/AltKit

## Discussion

First, I'm having compilation issues on my setup (macOS 12 Beta, XCode 13 Beta) with missing IOKIt import, so this code is untested.

This PR probably isn't the way the devs would implement this feature, but I created this to get it on the radar if the dev's want to use it as a starting point.


JoeMatt:
Quick update. Riley has sent me updated details to add some changes to this PR. It's still a draft though until I add that and they do an official release and I can change the swift module url from my tagged branch to their official master once they add a tag to fix swift pm not pulling a tag less branch (even if telling it to ignore release tags swiftpm is buggy...)

JoeMatt:
Rebased from master f7ece1da41675dfa5a62bf7e71c20b59824794c8 and switched the AltKit SPM repo from my fork to official @rileytestut repo

PR #3056, opened 2021-08-27
Use NSSavePanel on macOS for exporting

### LINKED ISSUE #3029: Improve Export Debug Log feature
The Debug Log feature is very helpful for diagnosing problems with UTM and guests. However it has two flaws: 

1. It is unclear what needs to be done to get a log, turning on the switch doesn't let the user know they have to start the VM again.

2. Clicking on the export button on macOS results in an iOS-like share sheet instead of a "Save as…" window like one might expect.

ktprograms:
### Summary of changes:
- Switch to using a NSSavePanel on macOS for exporting (log, QEMU command, .utm file)
- Don't copy debug log to temporary directory
- Add `appleLog` `UTType` extension (UTI of `com.apple.log`)
- Delete `SharingServicePicker.swift` and replace with `SavePanel.swift`

Fixes #3029


conath:
> I noticed the `project.pbxproj` has changes due to setting the `Sign to Run Locally` option.
> 
> But the changes relating to adding and removing files need to be committed. Is there a list of which parts in the `project.pbxproj` should be committed?

I don't have a list, but I have found in the past that there are large chunks of the file that remain unchanged, and therefore it is possible to just commit the parts that have the added files. Using a GUI for git makes this easier (e. g. GitHub Desktop, Sourcetree).

Platform/macOS/SavePanel.swift
```diff
@@ -0,0 +1,75 @@
...
+                    savePanel.allowedContentTypes = [.UTM]
+                }
+
+                savePanel.begin { result in
+                    if result == .OK {
+                        if let destUrl = savePanel.url {
+                            try? FileManager.default.copyItem(at: sourceUrl, to: destUrl)
+                        }
+                    } else {
+                        NSSound.beep()
```
conath, line 51: I'm not sure why you would beep when the result is not `ok`. Does this result in a beep when the user cancels the export dialog?

Platform/macOS/SavePanel.swift
```diff
@@ -0,0 +1,75 @@
...
+                } else if sourceUrl.pathExtension == "utm" {
+                    savePanel.title = "Select where to save UTM Virtual Machine:"
+                    savePanel.nameFieldStringValue = sourceUrl.deletingPathExtension().lastPathComponent
+                    savePanel.allowedContentTypes = [.UTM]
+                }
+
+                savePanel.begin { result in
+                    if result == .OK {
+                        if let destUrl = savePanel.url {
+                            try? FileManager.default.copyItem(at: sourceUrl, to: destUrl)
```
conath, line 48: Is an error available here? Errors can be shown in UTM by setting `UTMData.alertMessage`. (available as `@EnvironmentObject`)

Platform/macOS/SavePanel.swift
```diff
@@ -0,0 +1,75 @@
...
+                }
+            } else if let command = shareItem as? String {
+                savePanel.title = "Select where to export QEMU command:"
+                savePanel.nameFieldStringValue = "command"
+                savePanel.allowedContentTypes = [.plainText]
+
+                savePanel.begin { result in
+                    if result == .OK {
+                        if let destUrl = savePanel.url {
+                            try? command.write(to: destUrl, atomically: true, encoding: .utf8)
```
conath, line 62: see above comment re: error handling

Platform/macOS/SavePanel.swift
```diff
@@ -0,0 +1,75 @@
...
+                        }
+                    } else {
+                        NSSound.beep()
```
conath, line 65: see above comment re: beep on cancel

Platform/Shared/VMShareFileModifier.swift
```diff
@@ -24,7 +24,7 @@ struct VMShareItemModifier: ViewModifier {
     #if os(macOS)
     func body(content: Content) -> some View {
         ZStack {
-            SharingsPicker(isPresented: $isPresented, sharingItems: items())
+            SavePanel(isPresented: $isPresented, shareItem: items()[0]) // Only ever called with 1 item
```
conath, line 27: Can you change this to use a single value instead of an Array?

Platform/UTMExtensions.swift
```diff
@@ -97,6 +97,8 @@ extension View {
 extension UTType {
     // SwiftUI BUG: exportedAs: "com.utmapp.utm" doesn't work
     static let UTM = UTType(exportedAs: "utm")
+    
+    static let appleLog = UTType(filenameExtension: "log")!
```
conath, line 101: Is this needed when there is `UTType.log`?

Platform/macOS/SavePanel.swift
```diff
@@ -0,0 +1,75 @@
...
+    func updateNSView(_ nsView: NSViewType, context: Context) {
+        if isPresented {
+            let savePanel = NSSavePanel()
+            savePanel.directoryURL = URL(fileURLWithPath: "/Users/\(NSUserName())/Downloads")
+
+            if let sourceUrl = shareItem as? URL {
+                if sourceUrl.pathExtension == "log" {
+                    savePanel.title = "Select where to save debug log:"
+                    savePanel.nameFieldStringValue = "debug"
+                    savePanel.allowedContentTypes = [.appleLog]
```
conath, line 38: See comment to `UTMExtensions` change

UTM.xcodeproj/project.pbxproj
```diff
@@ -4322,7 +4322,7 @@
 					"@executable_path/../../../../../../../Frameworks",
 				);
 				MACOSX_DEPLOYMENT_TARGET = 11.0;
-				PRODUCT_BUNDLE_IDENTIFIER = "$(PRODUCT_BUNDLE_PREFIX:default=com.utmapp).QEMULauncher";
```
conath, line 4325: As mentioned in my comment to the PR, please drop the changes from this line onward, thanks.

conath (changes requested):
Thanks @ktprograms this is a welcome change. 😎
Please see my detailed feedback on the few changes I would recommend to this PR.

Platform/macOS/SavePanel.swift
ktprograms: Cancelling the dialog would make the result not `.OK`. The beep can and the `else` can be removed then there will be no audible feedback after cancelling the dialog.

Platform/Shared/VMShareFileModifier.swift
ktprograms: The other part of this (iOS and `UIActivityViewController`) needs an array. It's just that I didn't want to have handling for an array of items on macOS since the code never calls it with more than 1.

Platform/UTMExtensions.swift
ktprograms: `UTType.log` has a UTI of `public.log`. Console log files have a UTI of `com.apple.log`.

https://developer.apple.com/documentation/uniformtypeidentifiers/uttype/3551527-log

```
% mdls -name kMDItemContentType Library/Containers/com.utmapp.UTM/Data/Documents/Ubuntu\ 20.04\ arm\ 2.utm/debug.log 
kMDItemContentType = "com.apple.log"
```

If you don't specify a 'bottom of the tree' type, NSSavePanel ignores the `allowedContentTypes` and allows all types.

UTM.xcodeproj/project.pbxproj
ktprograms: Done in https://github.com/utmapp/UTM/pull/3056/commits/952eb1eb555a6da933864771fce4faae0896317d (force pushed to not mess up commit history)

ktprograms:
@conath I removed the `NSSound.beep()` calls since nowhere else in UTM beeps.

ktprograms:
@osy @conath The `SavePanel` code is a bit of a hack. The code checks if the `shareItem` is a `URL` or a `String`, then if it's a `URL` it checks if the file extension is `.log` or `.utm`, and changes behaviour based on that. If `shareItem` is a different type or has a different file extension it will just be ignored.

Is that ok or is there a better way to do it?

Platform/Shared/VMShareFileModifier.swift
conath: That makes sense. It would be good to mention this in a code comment.

Platform/UTMExtensions.swift
conath: TIL

Platform/Shared/VMShareFileModifier.swift
ktprograms: @conath Err, the comment is right on line 27

conath (changes requested):
Mostly fine now, see Comment on the PR

conath:
The better way to do it would be an enum with [different associated types](https://www.programiz.com/swift-programming/associated-value-enum#multiple-values) for the cases. For example:

```swift
enum ShareItem {
    case utmFile(URL)
    case plaintext(String)
}
```

Not sure, but you could then probably use [`if case let`](https://fuckingifcaseletsyntax.com/) to replace the [`if let`](https://github.com/utmapp/UTM/blob/aa0e9a34de989897e5de2e48a639db3ee1e8ec0c/Platform/macOS/SavePanel.swift#L35) you have currently.

ktprograms:
@conath should be good now

conath (commented):
Looks good to me! I have not tested this yet.

Issue #2688, opened 2021-07-29
snapshots / disposable VMs


ThomasWaldmann:
I'm used to use Qubes OS on my other machine, especially using disposable VMs for random web surfing or potentially dangerous stuff (like working with files from untrusted sources).

A disposable VM is a VM created from a given state (template VM), executes normally, but when shut down, all changes get thrown away. E.g. if you do web surfing, you do not have to worry about getting tracked long-term by cookies or similar mechanisms - or if you try installing random software just to have a quick look, you do not have to worry about spoiling your system somehow, because all state will get thrown away automatically at shutdown time. Also malware does not have a chance to persist.

Such a dispVM could be implemented with **qemu temporary snapshots**.

Normal snapshots would be also useful as a safety net, e.g. before running OS upgrades or critical SW installations inside your VMs. If anything goes wrong, one can just revert to a known-good state in a snapshot.

ThomasWaldmann:
BTW, oracle virtualbox has a GUI for managing snapshots.

juanandresgs:
Looks like #3067 is inching us closer. Any chance we can get a snapshot management view? This would be hugely valuable.
Willing to donate to support dev.

ktprograms:
Hi @juanandresgs, what do you mean by a snapshot management view?

BTW the PR is only for [temporary snapshots](https://wiki.qemu.org/Documentation/CreateSnapshot#Temporary_snapshots), not qcow2 snapshots with backing files.

ThomasWaldmann:
@ktprograms if you have virtualbox installed somewhere, just have a look at its snapshots dialogues.

One can create (permanent) snapshots there, go back to the state of some snapshot, delete/merge snapshots, etc.

ktprograms:
I just realised, QEMU seems to have 3 kinds of snapshots.

1. Using the `-snapshot` option for a temporary disposable VM.
2. `qcow2` backing files to have multiple images based on a single 'parent' image.
3. `qcow2` save states which you can create and revert back to using the `qemu-img` `snapshot` subcommand.

If I understand it, you would like (3) to be implemented, right?

ThomasWaldmann:
Well, guess we need 1 and 3.

2 would be also nice to have as a space saver maybe.

ktprograms:
[...]

PR #3067, opened 2021-08-30
Add support for running VM as temporary snapshot

ktprograms:
#2688 

I tried adding the `-snapshot` option in the `QEMU` settings and it works, so I've changed the UI to use it.

The `-snapshot` option is enabled when you click the Play button while holding the `command` key on macOS, and in a long press context menu on iOS 14.

I can change the UI if this design isn't wanted.

Also I'm not sure how to change the UI on legacy (11 - 13) iOS.


Platform/Shared/VMContextMenuModifier.swift
```diff
@@ -56,7 +57,13 @@ struct VMContextMenuModifier: ViewModifier {
                 } label: {
                     Label("Run", systemImage: "play.fill")
                 }
+                Button {
+                    data.run(vm: vm, runAsSnapshot: true)
+                } label: {
+                    Label("Run as Snapshot", systemImage: "play")
```
conath, line 63: Let's distinguish the snapshot button by using the `play.rectangle` symbol instead.

Platform/Shared/VMDetailsView.swift
```diff
@@ -93,6 +93,14 @@ struct Screenshot: View {
     let large: Bool
     @EnvironmentObject private var data: UTMData
     
+    @State private var commandDown = false
+    
+    #if os(macOS)
+    private var playImageName: String { commandDown ? "play.circle" : "play.circle.fill" }
```
conath, line 99: same as above; use `play.rectangle` for play-as-snapshot; meaning for filled use `play.rectangle.fill`

Also please change it to option instead of command key. In many apps incl. Finder you get alternative features from buttons while holding option/alt. ⌥

Platform/Shared/VMDetailsView.swift
```diff
@@ -111,12 +119,34 @@ struct Screenshot: View {
...
+                    data.run(vm: data.selectedVM!, runAsSnapshot: true)
+                } label: {
+                    Label("Run as Snapshot", systemImage: "play")
+                }
+                #endif
+            }
+            .onAppear {
+                #if os(macOS)
+                NSEvent.addLocalMonitorForEvents(matching: [.flagsChanged]) { event in
+                    commandDown = event.modifierFlags.contains(.command)
```
conath, line 145: please change to option/alt. ⌥

Platform/Shared/VMRunButton.swift
```diff
@@ -0,0 +1,86 @@
...
+
+@available(iOS 14, macOS 11, *)
+struct VMRunButton: View {
+    var padding: CGFloat
+    @EnvironmentObject private var data: UTMData
+
+    @State private var commandDown = false
+
+    #if os(macOS)
+    private var playImageName: String { commandDown ? "play" : "play.fill" }
```
conath, line 27: please use `play.rectangle`

Platform/Shared/VMRunButton.swift
```diff
@@ -0,0 +1,86 @@
...
+//
+
+import SwiftUI
+
+@available(iOS 14, macOS 11, *)
+struct VMRunButton: View {
+    var padding: CGFloat
+    @EnvironmentObject private var data: UTMData
+
+    @State private var commandDown = false
```
conath, line 24: change to option (see above)

Platform/Shared/VMRunButton.swift
```diff
@@ -0,0 +1,86 @@
...
+@available(iOS 14, macOS 11, *)
+struct VMRunHelpModifer: ViewModifier {
+    @Binding var commandDown: Bool
+
+    func body(content: Content) -> some View {
+        #if os(macOS)
+        content.help(commandDown ? "Run as Snapshot" : "Run selected VM")
+            .onAppear {
+                NSEvent.addLocalMonitorForEvents(matching: [.flagsChanged]) { event in
+                    commandDown = event.modifierFlags.contains(.command)
```
conath, line 70: same as above (⌥ )

conath (changes requested):
Looks good overall! This could be a useful feature IMO. 
Please take a look at my comments, esp. changing from the command key to option key. Thank you!

If you need help testing this, or with something else, let me know.

ktprograms:
BTW some of this code (the parts with custom modifiers) should probably be rewritten when Swift 5.5 is available with it's `#if`/`#endif` on modifiers.

ktprograms:
@conath I think I'll need some help adding the Run as Snapshot option for older iOS versions of the app.

I'm not sure how the UI should be implemented and where the code for it should be.

conath (approved):
Nice! @osy this looks like it will be a great addition.

ThomasWaldmann:
Thanks for working on this! See also #2688.

As a remark: I find the terms / language for this option a bit unusual (but of course that began with the qemu option name) and not very intuitive / understandable.

Snapshot usually means something frozen, something one is keeping. But here (if I understood it correctly), it rather means to run the VM as a "disposable VM", so everything one changes while running it is disposed as soon as it is shutdown. Qubes OS calls this a "disposable VM".

But also if one considers the perspective of a non-admin, non-developer user: they likely would not associate "run as snapshot" with "disposable" or "temporary / throw away".

Also, hiding this behind an additional key to hold down is not very intuitive / discoverable - would anybody even find this without reading the docs (are there docs?)?

ktprograms:
The holding down the key would probably be in the release notes when it's included, but I do agree that 'snapshot' isn't a very accurate name for it. As you said, it's because `qemu` uses the `-snapshot` flag to enable disposable VMs. Also, the `Run as Snapshot` option is available in the right click/long press context menu.

conath:
Good point @ThomasWaldmann. There should definitely be some kind of explanation whenever this feature is activated, possibly an alert explaining what will happen, with the option to run normally and to not show it again. The running VM window should probably also gain an indicator that changes won't be saved.

I'm not sure on the wording either, "Run without saving changes" could be a more verbose and more obvious name (I can't think of a concise way of phrasing it). Though if someone is looking for the QEMU feature "snapshot" then they wouldn't find it.

In terms of UI, as ktprograms mentioned there is the context menu action so it's discoverable. Maybe a menu bar item could be added as well, if there isn't already one included with the PR. The menu bar has the advantage of being searchable from the Help menu.

conath:
Maybe this feature could have similar UX to a web browser's incognito window where the history is deleted when the browser window is closed. There could be a little ghost icon in the toolbar whenever a VM is running in this "snapshot" mode.

ktprograms:
When using the Qemu Monitor, it's possible to commit changes to disk even if it was started as `-snapshot`. Does UTM have access to the Qemu Monitor?

Managers/UTMQemuSystem.m
```diff
@@ -645,6 +645,10 @@ - (void)argsFromConfiguration {
     // fix windows time issues
     [self pushArgv:@"-rtc"];
     [self pushArgv:@"base=localtime"];
+    
+    if (self.configuration.runAsSnapshot) {
```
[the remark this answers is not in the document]
adespoton: Agreed; this also makes it consistent with other VM systems.  On some UIs though, there's the start button > with a drop-down button on press-and-hold that lets you select between Run, Run Snapshot, Run Shadow Copy.
Of course, the issue there is that then you get into shadow copy management, where you need the ability somewhere to manage/delete shadow copies.

As for UI labeling, Snapshot is definitely problematic, especially when most popular VMs use the term the way QEMU refers to shadow copies.  Even the QEMU documentation has to differentiate between -snapshot and VM Snapshots: https://qemu.readthedocs.io/en/latest/system/images.html

On the plus side, that documentation now has a clear explanation of what VM snapshotting and -snapshot do, which might be useful to incorporate into UTM documentation as well.  Of course, with no access to the monitor, there's currently no way to manage VM snapshots on UTM?

Mark me down as someone who would love the ability to use -snapshot, plus the ability to optionally merge the snapshot into the image at the end of the session.  I do lots of demonstration sessions on various OSes, and would love the ability to automatically discard changes unless I explicitly choose to merge them.  A dialog when closing the VM stating "This session has unsaved changes.  Would you like to save them?" with UTM handling the response by doing nothing if the answer is No, or sending "<ctrl>-a s" to the serial console if the answer is Yes.

ktprograms:
@osy Done your requested changes (completely rewrote it against current master)

Managers/UTMQemuVirtualMachine.m
```diff
@@ -299,6 +299,15 @@ - (void)_vmStartWithCompletion:(void (^)(NSError * _Nullable))completion {
         }
     }
     
+    if (self.isRunningAsSnapshot) {
+        self.system.runAsSnapshot = self.isRunningAsSnapshot;
+    } else {
+        // Loading save states isn't possible when -snapshot is used
```
conath, line 305: > Loading save states isn't possible when -snapshot is used

Should this be exposed to the user somehow?

ktprograms: This was discussed in Discord, specifying "Run without saving changes" is an explicit user choice so it overrides the loading of a save state.

Do you mean there should be a warning that the save state will be lost?

conath: Yes, that’s what I was going for.

ktprograms: Anddddd it turns out my brain isn't running correctly...

This is basically just a safeguard, but the context menu UI does not show the "Run" or "Run without saving changes" button when there is a save state.

idk how I completely forgot about it and spent 15 mins implementing an alert... :(

Managers/UTMQemuSystem.m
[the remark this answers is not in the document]
osy: @adespoton I plan to merge the PR as-is but your suggestions are good. Feel free to open a new issue with the suggestions once it's merged.

osy (approved):

osy (approved):

PR #3080, opened 2021-09-04
Fix cropped text in Config Display tab

ktprograms:
For some reason, the `Requires SPICE guest agent tools to be installed. Retina Mode is recommended only if the guest OS supports HiDPI.` text wouldn't wrap unless I added `.frame(height: 32)` to force it to be the right height.

Hardcoding the height seems ok to me since the Config window has a minimum width so the text won't be wrapped to more than 2 lines long.


conath (changes requested):
Could this be fixed instead by moving the Text() out of the footer property? I'd rather not change the order of the settings to work around what looks like a SwiftUI bug.

ktprograms:
@conath even moving the text out of the footer didn't work, but [this](https://stackoverflow.com/questions/56505929/the-text-doesnt-get-wrapped-in-swift-ui/59277022#59277022) did.

Also had to add horizontal padding on macOS only.

I force pushed and overwrote the previous commits since I needed to start over anyway.

conath (approved):

PR #3121, opened 2021-09-20
Add virtio-rng-pci device to qemu commandline

ktprograms:
https://wiki.qemu.org/Features/VirtIORNG#Invocation

@conath could you help to test that this doesn't break Windows etc. VMs? You can just add `-device virtio-rng-pci` to the QEMU arguments.

**Edit:**
It seems according to [this](https://www.redhat.com/en/blog/red-hat-enterprise-linux-virtual-machines-access-random-numbers-made-easy), virtio RNG should work on Windows guests too?

### DIFF
--- a/Managers/UTMQemuSystem.m
+++ b/Managers/UTMQemuSystem.m
@@ -638,6 +638,10 @@ static size_t sysctl_read(const char *name) {
             [self pushArgv:@"-device"];
             [self pushArgv:self.configuration.displayCard];
         }
+        if (self.configuration.systemBootUefi) {
+            [self pushArgv:@"-device"];
+            [self pushArgv:@"virtio-rng-pci"];
+        }
     }
 }
 


conath:
What is this for?

ktprograms:
I have been seeing this in the boot process (the `no randomness supplied` part):
Screenshot 2021-09-20 at 2 45 16 PM

When I saw the QEMU documentation on `virtio-rng-pci` I tried adding it to the QEMU arguments and it worked.
This is what it looks like with adding `-device virtio-rng-pci`:
Screenshot 2021-09-20 at 2 48 54 PM

conath (changes requested):
I will test this later on but I already see a compatibility issue: I believe this should only be enabled when EFI boot is enabled because older guests will not be able to interface with this.

ktprograms:
> older guests will not be able to interface with this

Is it only available for EFI? The `virtio-rng-pci` device just provides a virtio compliant source of randomness (https://wiki.qemu.org/Features/VirtIORNG). The boot messages are just the immediately visible change.

conath:
What I meant by old guest is ISA only systems or systems that don’t use modern PCI devices. The documentation reads as if it will always add a PCI device. (Maybe it should just be disabled for `pc` system?)

You can leave it like this until I get a chance to test an older system or you can test yourself with a `pc` and `ppc` system. Thanks.

conath (changes requested):
I tested both the `-device virtio-rng` and `-device virtio-rng-pci` variants and they both cause error on QEMU start with ISA-only systems. So there needs to be a check if the emulated system has a PCI bus.

osy:
It can be gated by `self.configuration.systemBootUefi`

ktprograms:
@osy Sorry, this PR wasn't one of my top priorities, but I think it should be good now.

conath (approved):

PR #3154, opened 2021-10-03
Show alert before closing window or quitting

### LINKED ISSUE #3103: Closing VM window with the title bar close button doesn't show confirmation dialog
**Describe the issue**
When closing a VM window using the title bar close button doesn't result in the confirmation dialog being shown. I would expect to confirmation dialog to be shown similar to when you press Command + W.

**Configuration**
* UTM Version: 2.2.2
* Intel or Apple Silicon? Apple Silicon

ktprograms:
Fixes #3103 by putting the alert code in windowShouldClose/applicationShouldTerminate instead of listening for key press events

I'm intentionally using the blocking `runModal` on the Alerts since you wouldn't want to be able to ignore a `Do you want to quit` alert.

<sub><sub>Side note: Isn't it funny how Swift 5.5 is available on Mac but the new APIs like `Task` aren't? For example I can use `#if` on modifiers already...</sub></sub>

### DIFF
--- a/Platform/UTMApp.swift
+++ b/Platform/UTMApp.swift
@@ -26,6 +26,11 @@ struct UTMApp: App {
     var body: some Scene {
         WindowGroup {
             ContentView().environmentObject(data)
+            #if os(macOS)
+                .onAppear {
+                    appDelegate.data = data
+                }
+            #endif
         }.commands { VMCommands() }
         #if os(macOS)
         Settings {
--- a/Platform/macOS/AppDelegate.swift
+++ b/Platform/macOS/AppDelegate.swift
@@ -15,7 +15,39 @@
 //
 
 class AppDelegate: NSObject, NSApplicationDelegate {
+    var data: UTMData?
+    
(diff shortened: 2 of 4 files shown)


ktprograms:
BTW @osy can you please update the GH Actions to use Xcode 13. And when you do that I think you'll need to change the `#if swift(>=5.5)` to something else since the new APIs still aren't available on macOS even though Swift 5.5 is.

osy:
> And when you do that I think you'll need to change the `#if swift(>=5.5)` to something else since the new APIs still aren't available on macOS even though Swift 5.5 is.

That’s exactly why we switched back to Xcode 12 https://github.com/utmapp/UTM/commit/171abc2aedc2a728e1922eabdbc666bdeeb272dd

ktprograms:
> That’s exactly why we switched back to Xcode 12 171abc2

So you won't use Xcode 13 until Monterey is released?

osy:
Yes because it’s impossible to know if it’s Xcode 13 beta which has macos 12 sdk or Xcode 13 gm which has macos 11 sdk.

PR #3155, opened 2021-10-03
Differentiate the main UTM window from VM windows

### LINKED ISSUE #3099: Differentiate the main UTM window title from VM windows
After clicking Play to run a VM on macOS, the main UTM window and the VM's window have the same title:

Screen Shot 2021-09-10 at 12 50 12

This is confusing when trying to switch between them using the menu bar item "Window".
The VM and main windows should each gain additional text to indicate which one is which. For example, the VM window could indicate the state of the VM ("Windows 10 installed – running").

ktprograms:
Fixes #3099 by setting isExcludedFromWindowsMenu, then adding a custom
'UTM' menu item that brings the main UTM window front

Note: Should the `UTM` menu item have a different name (like `Open Main Window`) since just `UTM` isn't clear what it does?

### DIFF
--- a/Platform/Shared/VMCommands.swift
+++ b/Platform/Shared/VMCommands.swift
@@ -19,6 +19,9 @@ import SwiftUI
 @available(iOS 14, macOS 11, *)
 struct VMCommands: Commands {
     @Environment(\.openURL) private var openURL
+    #if os(macOS)
+    var appDelegate: AppDelegate
+    #endif
     
     @CommandsBuilder
     var body: some Commands {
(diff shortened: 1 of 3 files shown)


osy:
@ktprograms is this still needed after #3125 is merged? The main window will be named "UTM (VM name here)" in the menu.

osy:
Right, but this seems really fragile and outside of SwiftUI. My hope is to avoid doing things outside of SwiftUI when possible for maintainability. It seems like the immediate issue of unable to differentiate the main window would be fixed with a subtitle.

ktprograms:
This is what it shows up as for me. So not ideal since the VM name is less prominent, but I guess can't be helped if you want to stick with SwiftUI.

Screenshot 2021-10-08 at 8 44 50 AM

Can probably close this PR then

PR #3216, opened 2021-10-27
Set USB-XHCI bus for USB storage when using aarch64 system architecture

### LINKED ISSUE #3194: Windows Installer triggers QEMU USB reset, aborting installation
When trying to install Windows 11 ARM version, during the second step of the installation (Getting files ready for installation), this randomly ends with error 0x800701B1 (Windows cannot install required files. Make sure all files required for installation are available.) 

0-phase during which this is experienced
1-error seen during install

This seems to happen at random moments during this phase, somewhere between 0 and 100%. A single time, without any specific modifications to the settings, I managed to get things going to 100%. After that, the rest of the install worked as expected and so did the installed Windows version.

When trying to restart setup, without a VM restart, I noticed that the CD/DVD-drive is no longer readable. Changing the ISO does not seem to help. 

2-re-trying install after error
3a-cmd after error was seen
3b-cmd after error was seen

**Configuration**
* UTM Version: 2.2.4 (36)
* OS Version: 11.6
* Intel or Apple Silicon? M1 Silicon (Macbook Air 2020)

4-settings

**Debug log**
[debug.log](https://github.com/utmapp/UTM/files/7374458/debug.log)

conath:
~~Fixes~~ Partially fixes #3194 (Windows 11 installer causes QEMU EHCI reset, aborting install)

The result of this PR is that for `aarch64` guest, the USB CD drive is connected to the XHCI bus:
Screen Shot 2021-10-27 at 22 36 55

This has the nice side effect of increasing the speed of the USB drive.

Edit: added "partially" above to reflect my findings while testing. See [my comment](https://github.com/utmapp/UTM/issues/3194#issuecomment-953314235) for details.

### DIFF
--- a/Managers/UTMQemuSystem.m
+++ b/Managers/UTMQemuSystem.m
@@ -209,7 +209,10 @@ static size_t sysctl_read(const char *name) {
         [self pushArgv:[NSString stringWithFormat:@"nvme,drive=%@,serial=%@,bootindex=%lu", identifier, identifier, bootindex++]];
     } else if ([interface isEqualToString:@"usb"]) {
         [self pushArgv:@"-device"];
-        [self pushArgv:[NSString stringWithFormat:@"usb-storage,drive=%@,removable=%@,bootindex=%lu", identifier, removable ? @"true" : @"false", bootindex++]];
+        /// use usb 3 bus for virt system, unless using legacy input setting (this mirrors the code in argsForUsb)
+        bool useUSB3 = !self.configuration.inputLegacy && [self.configuration.systemTarget hasPrefix:@"virt"];
+        NSString *bus = useUSB3 ? @",bus=usb-bus.0" : @"";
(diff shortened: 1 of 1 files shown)


conath:
That could work. Why not always generate the XHCI bus regardless of the input setting?

osy:
Lots of machines don’t support it. And machines that do may not support it in the OS.

conath:
I'm thinking we could gate it behind the EFI boot flag or specific systems. Or maybe there should be a USB2/3 switch like in other apps.

The easiest is probably your suggestion 😃

conath:
should be fine now

lonione:
Faced the same issue with 2.4.1 and Windows 11 22533.1001_PROFESSIONAL_ARM64_EN-US.ISO 
Quite frustrating, as it kept failing (quite early, maybe around 5%) even after changing the XHCI bus. However, now it has successfully completed with the same config - so I can confirm it works eventually but requires some patience and persistence :) Thanks for the information in this thread.

PR #3226, opened 2021-10-30
Fix downloadVM URL action failing to import downloaded file

conath:
A file path confusion existed where the downloaded ZIP was extracted to the UTM
Documents directory instead of the temporary folder (I believe this change may have been caused by an update of the ZIP library in use). As a result, UTM failed to import the downloaded file because a folder with the same name already existed (as seen in the console log).

This is fixed by manually specifying the path for the unzip call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,9 @@ UTM is also available for macOS: https://mac.getutm.app/
 
 UTM is distributed under the permissive Apache 2.0 license. However, it uses several (L)GPL components. Most are dynamically linked but the gstreamer plugins are statically linked and parts of the code are taken from qemu. Please be aware of this if you intend on redistributing this application.
 
-Some icons made by [Freepik](https://www.freepik.com) from [www.flaticon.com](https://www.flaticon.com/).
-
+Some icons made by [Freepik](https://www.freepik.com) from [www.flaticon.com](https://www.flaticon.com/).  
+UTM uses [ZIP Foundation](https://github.com/weichsel/ZIPFoundation), which is released under the MIT License.
+  
   [1]: https://github.com/utmapp/UTM/actions?query=event%3Arelease+workflow%3ABuild
   [2]: screen.png
   [3]: https://github.com/ktemkin/qemu/blob/with_tcti/tcg/aarch64-tcti/README.md
--- a/Platform/Shared/UTMImportFromWebTask.swift
+++ b/Platform/Shared/UTMImportFromWebTask.swift
@@ -46,12 +46,13 @@ class UTMImportFromWebTask: NSObject, URLSessionDelegate, URLSessionDownloadDele
                 try fileManager.removeItem(at: downloadedZip)
             }
             try fileManager.moveItem(at: location, to: downloadedZip)
-            let unzippedURL = try Zip.quickUnzipFile(downloadedZip)
+            let unzippedPath = tempDir.appendingPathComponent(originalFilename.replacingOccurrences(of: ".zip", with: ""))
+            try Zip.unzipFile(downloadedZip, destination: unzippedPath, overwrite: true, password: nil)
             /// remove the downloaded ZIP file
             try fileManager.removeItem(at: downloadedZip)
-            handleUnzipped(unzippedURL)
+            handleUnzipped(unzippedPath)
             /// remove unzipped file
-            try FileManager.default.removeItem(at: unzippedURL)
+            try FileManager.default.removeItem(at: unzippedPath)
         } catch {
             logger.error(Logger.Message(stringLiteral: error.localizedDescription))
             try? fileManager.removeItem(at: downloadedZip)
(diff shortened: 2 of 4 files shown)


osy:
@conath maybe it’s better to extract it to the Inbox. That way importUTM moves it instead of copying it and you don’t need 3x the space to download the zip, extract it, copy it, then delete the copy?

conath:
With APFS on iOS and Mac the copying will basically be a no-op and won't count as additional used space (IIRC), so I don't see this as a problem.

If you still want to change it to the Inbox, how do I get the path to the Inbox folder?

osy:
> won't count as additional used space 

Are you sure that’s true?

Anyways I think Inbox won’t work on macos so it’s not a good solution. Can extract the .utm directly to Documents?

conath:
Yes, I could do it straight to Documents. I would change the extracted path to the Documents and make sure the resulting folder doesn't have the same name as the VM (e. g. add UUID), then I would keep the existing logic that finds the actual .utm file in the folder and move that to the Documents folder.

conath:
Cloning instead of copying is one of the main features of APFS: [About Apple File System: Clones Reduce the Cost of Copying](https://developer.apple.com/documentation/foundation/file_system/about_apple_file_system#2985645)

osy:
> Cloning instead of copying is one of the main features of APFS

Yes, but it still counts as disk space used afaik.

> I would change the extracted path to the Documents and make sure the resulting folder doesn't have the same name as the VM

Why can’t you extract it in place? The ZIP doesn’t have an “outer” folder wrapping the .utm. If you extract it directly with the command line on macos you’ll see only the .UTM file.

conath:
The result of unzipping with this Zip library is a folder with the name of the zip, minus .zip, and inside is the actual UTM package. So there is an outer folder, which is why I added the extra step of finding the contained .utm in the first place. Maybe we could check what’s in the ZIP and only extract the .utm?

osy:
Yeah that sounds good.

conath:
Success! Implementing the partial-unzip took more work than anticipated, as I had to switch out the Zip library. The new code also includes a duplicate name avoidance feature, please see the full commit message. https://github.com/utmapp/UTM/pull/3226/commits/8185c2e065bef62bfc87224c67ec08e8af3fc95c

conath:
@osy the ZipFoundation library is MIT licensed. (Where) do we show the open source licenses in the app?

osy:
http://apache.org/legal/resolved.html says we can use it in an Apache licensed project.

At the end of README.md, I say we add

`[ZIP Foundation](https://github.com/weichsel/ZIPFoundation) is released under the MIT License.`

conath:
Sure, can do that. The license should also be added to the shipping app somehow, so I believe it would make sense if you add it to the bottom of the [Licenses](https://mac.getutm.app/licenses/) page, as that is linked to from the app. I'm not sure if this link is available on iOS. If it is not, we should address that soon. That page also seems to be more up to date than the License file in the repo.

osy:
Please open an issue about that so I don’t forget.

conath:
Not sure why it CI failed. I had cancelled the run for the README change, that's all. I rebased the changes on master now and it builds fine on my machine.

README.md
```diff
@@ -56,3 +56,5 @@ Some icons made by [Freepik](https://www.freepik.com) from [www.flaticon.com](ht
   [3]: https://github.com/ktemkin/qemu/blob/with_tcti/tcg/aarch64-tcti/README.md
   [4]: https://github.com/ish-app/ish
   [5]: https://github.com/holzschu/a-shell
+  
+[ZIP Foundation](https://github.com/weichsel/ZIPFoundation) is released under the MIT License.
```
osy, line 60: Can you move this above the links and below `Some icons...`? The way markdown is usually formatted is that you put references at the bottom of the document (unless they're inline).

conath:
It seems to fail to compile the IQKeyboardManager dependency. I am confused.

```log
/Users/runner/Library/Developer/Xcode/DerivedData/UTM-ezfhxplkjtgbtmeymgqkuzyfypfk/SourcePackages/checkouts/IQKeyboardManager/IQKeyboardManagerSwift/IQToolbar/IQBarButtonItem.swift:71:36: error: cannot convert value of type 'String' to expected argument type 'NSAttributedString.Key'
392
                    textAttributes[key] = value
393
                                   ^
394
                                   NSAttributedString.Key(rawValue:  ) ?? <#default value#>
```

conath:
Interesting. I can reproduce the build failure if I update the IQKeyboardManager package from version 6.5.6 to 6.5.7. Limiting it to 6.5.6 gets rid of the compile errors.

conath:
This commit might be what's breaking the build. https://github.com/hackiftekhar/IQKeyboardManager/commit/709a01dbb7955f80045578392548836fca6ca2c9

osy:
Oh I didn't realize it was set to `upToNextMajorVersion`. Good catch.

Also bad on them for making API breaking changes without incrementing the major version. Like wtf.

PR #3227, opened 2021-10-31
Add support for Disk Images on external Drives (macOS only)

### LINKED ISSUE #3138: The public availability of VMware Fusion for Apple Silicon Public Tech Preview can directly use external virtual hard disk.
**Describe the issue**
v2.2.4 cannot directly use virtual machines on my external SSD. When I open the virtual machines it copies the whole file to the internal SSD but it lacks so much storage to handle VM(100GB+).
I download and install the [VMware-Fusion-e.x.p-18656771_arm64.dmg](https://download3.vmware.com/software/fusion/file/VMware-Fusion-e.x.p-18656771_arm64.dmg). In the disk share mode it can use the virtual hard disk on my external SSD directly.

I'm waiting for it.

**Configuration**
* UTM Version: v2.2.4
* OS Version: 11.6
* Apple Silicon

ktprograms:
#### ToDo:
- [x] Fix the bookmark file not being moved to the final directory (currently it only works if you have an already created VM).
- [x] Opening the Drives configuration pane initially shows both the drive and the drive bookmark. And later the bookmark becomes deleted.
~~These above 2 problems are because I store the bookmark data in a file with the image name + `.bookmark` extension, so the orphaned drives check is considering it orphaned. I **can** store the name with the `.bookmark` extension in the config but then users will see (for example) `disk-0.qcow2.bookmark` in the Drives tab. The other alternative is for the Drives tab to remove the `.bookmark` extension before putting it into the UI.~~
- [ ] Show error alert if bookmark cannot be resolved (if the drive isn't plugged in). Bit difficult since resolving the bookmark is done in `UTMQemuSystem.m` which doesn't have access to the UI. (@osy some help here please?)
- [ ] In the `VMConfigDriveDetailsView` show a text with which volume name the bookmark is from (might not be possible since the bookmark is a NSData and it can't be resolved into a URL if the drive folder doesn't exist).
- [x] In the `VMConfigDriveDetailsView` add a button to move a VM drive to an external drive
- [x] Add ability to create new disk in external drive

Fixes #3138 and #3142

The changes to `UTM.xcodeproj/project.pbxproj` are alphabetically re-ordering the files conath added in the URL scheme PR.

### DIFF
--- a/Platform/UTMData.swift
+++ b/Platform/UTMData.swift
@@ -468,7 +468,8 @@ class UTMData: ObservableObject {
         }
     }
     
-    // MARK: - Import Disk Image from External Drive
+    // MARK: - External Drive Disk Image Functions
+
     func importDriveFromExternal(_ image: URL, for config: UTMConfiguration) throws {
         _ = image.startAccessingSecurityScopedResource()
         defer { image.stopAccessingSecurityScopedResource() }
@@ -507,6 +508,31 @@ class UTMData: ObservableObject {
                 interface = "none"
             }
             config.newDrive(name, path: path, type: imageType, interface: interface, bookmark: true)
+        }
+    }
+
+    func moveDriveToExternal(at index: Int, url dstUrl: URL, for config: UTMConfiguration) throws {
+        // Move the Disk image to dstUrl
+        guard let srcPath = config.driveImagePath(for: index) else {
+            return
+        }
+        let srcUrl = URL(fileURLWithPath: srcPath, relativeTo: config.imagesPath)
+        do {
+            if fileManager.fileExists(atPath: dstUrl.path) {
+                try fileManager.removeItem(at: dstUrl)
+            }
+            try fileManager.copyItem(at: srcUrl, to: dstUrl)
+            try fileManager.removeItem(at: srcUrl)
+            config.removeDrive(at: index)
+        } catch {
+            throw NSLocalizedString("Failed to move drive to external disk: \(error.localizedDescription)", comment: "UTMData")
+        }
+
+        // Now import the drive from the external disk
+        do {
+            try importDriveFromExternal(dstUrl, for: config)
+        } catch {
+            throw NSLocalizedString("Failed to import drive back from external disk: \(error.localizedDescription)", comment: "UTMData")
         }
     }
 
--- a/Platform/UTMData.swift
+++ b/Platform/UTMData.swift
@@ -470,7 +470,7 @@ class UTMData: ObservableObject {
     
     // MARK: - External Drive Disk Image Functions
 
-    func importDriveFromExternal(_ image: URL, for config: UTMConfiguration) throws {
+    func importDriveFromExternal(_ image: URL, for config: UTMConfiguration, interface: String? = nil) throws {
         _ = image.startAccessingSecurityScopedResource()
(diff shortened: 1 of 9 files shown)


ktprograms:
@osy can you please take a look at this?

conath:
I have a suggestion regarding the error handling: there is already some error handling in place in [- (void)startWithCompletion](https://github.com/utmapp/UTM/blob/466fa95453cfd4b10f532d40b1cd090c67ea8405/Managers/UTMQemuSystem.m#L787). You could add an error parameter to `updateArgvWithUserOptions` and pass it through the chain to `argsForDrives`.

Platform/UTMData.swift
```diff
@@ -468,6 +468,113 @@ class UTMData: ObservableObject {
...
+    func importDriveFromExternal(_ image: URL, for config: UTMConfiguration, interface: String? = nil) throws {
+        _ = image.startAccessingSecurityScopedResource()
+        defer { image.stopAccessingSecurityScopedResource() }
+
+        let path = image.appendingPathExtension("bookmark").lastPathComponent
+
+        var isDir: ObjCBool = false
+        guard fileManager.fileExists(atPath: image.path, isDirectory: &isDir), !isDir.boolValue else {
+            if image.pathExtension == "utm" {
+                throw NSLocalizedString("You cannot import a .utm package as a disk image. Did you mean to open the package with UTM?", comment: "UTMData")
```
conath, line 482: Why are you throwing Strings? If you need this to be a public function I would expect it to use the `self.alertMessage` like you did further down. Otherwise, please consider changing this and similar functions to `private`.

ktprograms: If you look elsewhere in UTMData the code also throws Strings because calls to UTMData functions are done in a `data.busyWork` block which catches exceptions and sets the alert message. The manually setting alertmessage was only inside the SavePanel closure because you can't throw from there into the caller.

osy:
Heads up the dev-monterey branch will be merged at some point so it may make sense to work from on top of it.

ktprograms:
> Heads up the dev-monterey branch will be merged at some point so it may make sense to work from on top of it.

I'd prefer to have to fix conflicts when merging `dev-monterey` but leave new work on `master` so these features can be available before the merge. Also `dev-monterey` is very far behind with the commits.

ktprograms:
> I have a suggestion regarding the error handling: there is already some error handling in place in [- (void)startWithCompletion](https://github.com/utmapp/UTM/blob/466fa95453cfd4b10f532d40b1cd090c67ea8405/Managers/UTMQemuSystem.m#L787). You could add an error parameter to `updateArgvWithUserOptions` and pass it through the chain to `argsForDrives`.

I'm not sure if that's ideal since it's a 3 layer deep chain, and also `updateArgvWithUserOptions` is used in other places than `startWithCompletion`.

osy:
> > Heads up the dev-monterey branch will be merged at some point so it may make sense to work from on top of it.
> 
> I'd prefer to have to fix conflicts when merging `dev-monterey` but leave new work on `master` so these features can be available before the merge. Also `dev-monterey` is very far behind with the commits.

Sure I’m just giving a heads up. This won’t make it in before 3.0.

ktprograms:
Superseded by #3484

PR #3251, opened 2021-11-04
Fix dual sandbox path

Halo-Michael:
This change will fix an issues cause UTM got dual sandbox path on iOS.
When installing UTM, iOS will create an sandbox named the CFBundleIdentifier: _com.utmapp.UTM_. However, if we signed UTM main executable with ldid, the executable's Identifier will be as _UTM_! So iOS will make UTM running in another sandbox path that difference with the system records. This change will make ldid sign UTM's Identifier as _com.utmapp.UTM_.
DF6AD7B2-3F7E-48E0-B1C0-B25DCBA963F4


osy:
Thanks for figuring this out. I've made an update that allows UTM SE to be fake signed properly as well.

PR #3292, opened 2021-11-08
Add a documents website

Mar10Josh:
Documents website for extra docs and people who don't want to go to github website


osy:
Thanks but the problem is now I have to manually keep two copies to docs in sync and remember to update both. Maybe there's a better way using git submodules or something...

Mar10Josh:
> Thanks but the problem is now I have to manually keep two copies to docs in sync and remember to update both. Maybe there's a better way using git submodules or something...

About that, I don't know git submodules.
Or, if you were talking about in pages, It doesn't exist. It's only /(root) or /docs

osy:
Closing due to my comment above

PR #3318, opened 2021-11-17
iOS: Make it possible to allocate more than about 5 GB of memory

kkk669:
As described [here](https://developer.apple.com/forums/thread/685084?answerId=695278022#695278022), you need two entitlements in order to use more than 5 GB of memory on iOS and iPadOS.

`com.apple.developer.kernel.increased-memory-limit` already has been added to [iOS.entitlements](https://github.com/utmapp/UTM/blob/6524c733d36f183c5b393b54e91f7f0cd2ae7ad7/Platform/iOS/iOS.entitlements#L5), but it's insufficient.

The results of my experiments in my app showed that `com.apple.developer.kernel.increased-memory-limit` certainly increased the return value of `os_proc_available_memory`, it was about 12 GB on the iPad Pro that has 16GB of memory, but the actual limit of memory that can be allocated was only about 5 GB. So I also added the `com.apple.developer.kernel.extended-virtual-addressing` entitlement, and now the app can allocate about 12 GB.

Therefore, you should add this entitlement in order to use more than 5 GB of memory.

### DIFF
--- a/Platform/iOS/iOS.entitlements
+++ b/Platform/iOS/iOS.entitlements
@@ -4,5 +4,7 @@
 <dict>
 	<key>com.apple.developer.kernel.increased-memory-limit</key>
 	<true/>
+	<key>com.apple.developer.kernel.extended-virtual-addressing</key>
+	<true/>
 </dict>
 </plist>
--- a/scripts/package.sh
+++ b/scripts/package.sh
@@ -195,6 +195,8 @@ deb )
 	<true/>
 	<key>com.apple.developer.kernel.increased-memory-limit</key>
 	<true/>
+	<key>com.apple.developer.kernel.extended-virtual-addressing</key>
+	<true/>
 </dict>
 </plist>
 EOL
@@ -211,6 +213,8 @@ ipa )
 	<key>get-task-allow</key>
 	<true/>
 	<key>com.apple.developer.kernel.increased-memory-limit</key>
+	<true/>
+	<key>com.apple.developer.kernel.extended-virtual-addressing</key>
 	<true/>
 	<!-- https://siguza.github.io/psychicpaper/ -->
 	<!---><!-->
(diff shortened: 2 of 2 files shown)


kkk669:
It seems like the CI failed because the request to https://ftp.pcre.org/pub/pcre/pcre-8.37.tar.bz2 timed out. It's not related to this PR.

osy:
Should also be added in both places in https://github.com/utmapp/UTM/blob/master/scripts/package.sh for IPA and fake IPA

kkk669:
@osy Thank you. I fixed it.

osy (approved):

don17sch:
I had some problems getting this to work. 
I had to make one more modification to get this to work
https://github.com/utmapp/UTM/blob/master/Build.xcconfig

I had to change #include? "CodeSigning.xcconfig" on line 24 to #include "CodeSigning.xcconfig"

once I did that I no longer was restricted to only 5GB of RAM

last-Programmer:
> As described [here](https://developer.apple.com/forums/thread/685084?answerId=695278022#695278022), you need two entitlements in order to use more than 5 GB of memory on iOS and iPadOS.
> 
> `com.apple.developer.kernel.increased-memory-limit` already has been added to [iOS.entitlements](https://github.com/utmapp/UTM/blob/6524c733d36f183c5b393b54e91f7f0cd2ae7ad7/Platform/iOS/iOS.entitlements#L5), but it's insufficient.
> 
> The results of my experiments in my app showed that `com.apple.developer.kernel.increased-memory-limit` certainly increased the return value of `os_proc_available_memory`, it was about 12 GB on the iPad Pro that has 16GB of memory, but the actual limit of memory that can be allocated was only about 5 GB. So I also added the `com.apple.developer.kernel.extended-virtual-addressing` entitlement, and now the app can allocate about 12 GB.
> 
> Therefore, you should add this entitlement in order to use more than 5 GB of memory.

I have a developer account. What are all the steps i need to taken into account when i debug from xcode in iPad Pro M1 with 16gb. So that i can use more than 3072MB for a virtual machine.

Thanks

last-Programmer:
I was able to solve it by setting PAID_ACCOUNT = YES in .xcconfig.

Thanks for the awesome documentation by the team.

Issue #3308, opened 2021-11-12
ERROR: could not get https://ftp.pcre.org/pub/pcre/pcre-8.37.tar.bz2


Alex-Braine:
Hello,
https://ftp.pcre.org is down, so "./scripts/build_dependencies.sh -p macos -a arm64" breaks off...

conath:
In the mean time, if you are okay with using pre-built dependencies, you can use the `sysroot-macos-arm64` generated by GitHub Actions. https://github.com/utmapp/UTM/actions/runs/1430056582

kkk669:
According to https://pcre.org/:

> Note that the former ftp.pcre.org FTP site is no longer available. You will need to update any scripts that download PCRE source code to download via HTTPS, Git, or Subversion from the new home on GitHub instead.

kkk669:
Maybe we need to wait for the next release of glib.

https://github.com/GNOME/glib/commit/e5dc299701df7189685d322cb709649ce3c9c81d#diff-bcf487a428afba1739e08e831ec3906c27fc16a195459d76bccd2083ce0c72f3

Alex-Braine:
https://github.com/GNOME/glib/releases/tag/2.70.2

conath:
I've spent some time yesterday trying to fix the CI by upgrading some or all of the source dependencies. I learned quickly I am in way over my head and it didn't end up working. Maybe the patches need to be updated. See #3380 (update all) and #3381 (update glib)

Alex-Braine:
https://github.com/Alex-Braine/qemu/tree/48310571a6ca7e1a50a3c0ca71923d35f1d8e560/scripts -> https://github.com/Alex-Braine/qemu/actions/runs/1603384020
Changes only in scripts/build_dependencies.sh and patches/sources.
I'll try to solve the python's "six" problem in order to update spice-gtk.

PR #3334, opened 2021-11-24
Fix #3333 floppy drive is read only

### LINKED ISSUE #3333: Floppy drive is read-only
There is an oversight in the [code for removable drives](https://github.com/utmapp/UTM/blob/6524c733d36f183c5b393b54e91f7f0cd2ae7ad7/Managers/UTMQemuSystem.m#L309) where the media type of the drive is always generated as `cdrom` for removable drives, without checking if the drive is a floppy. This leads to read-only access in the guest.

conath:
Fixed by checking if the drive interface is "floppy"; if so the media type is set correctly to "disk" instead of "cdrom".


conath:
CI is failing due to #3308

osy:
Please rebase to fix failing CI

conath:
Done

PR #3379, opened 2021-12-09
l10n: zh_TW: update translation

pan93412:
With [`Use Compiler to Extract Swift Strings`](http://www.wwdcnotes.com/notes/wwdc21/10220/)


kev1bart (approved):

pan93412:
See #3440

PR #3408, opened 2021-12-25
Fix Microphone access (on macOS)

conath:
This so far only requests permission when saving a VM with HDA card, needs to be extended to cover AC97 as well. Preferably it should be determined by the VM Config if microphone access is needed, and not in the `UTMData`.

Also just giving access does not seem to fix the guest mic access issue (recording still contains silence).


osy:
Silence in recording was fixed in 4.0.0

PR #3409, opened 2021-12-25
Fix #3388 invalid drive id for manually placed files

### LINKED ISSUE #3388: New VM Launch Returns Error: Identifiers consist of letters… when placed image file in .UTM manually
**Describe the issue**
Error popup stating `Identifiers consist of letters, digits, '-', '.', '_', starting with a letter.` while trying to launch new VM.

New VM includes qcow2 file generated using qemu-img installed using brew. 

**Configuration**
* UTM Version: 2.4.1 (40)
* OS Version: 11.6.2
* Intel or Apple Silicon? Apple (M1)

**Crash log**
No crash reports in Console, no file names containing "qemu" in log reports section, either.

**Debug Log**
[debug.log](https://github.com/utmapp/UTM/files/7721628/debug.log)

**Upload VM**
[config.plist.txt](https://github.com/utmapp/UTM/files/7721632/config.plist.txt)

config.plist includes this:

```
	<key>Drives</key>
	<array>
		<dict>
			<key>DriveName</key>
			<string>7E113ED5-C4F0-4709-98AC-FD11309A7949</string>
```

It seems plausible this could be the same as closed issue #2614 for iOS.  In this case, can I edit that string in the plist file to start with a letter as a workaround?

conath:
QEMU drive ids must start with a letter.
However if an image files was manually placed in the VM images folder,
this code path is triggered and a UUID set for the id.
This causes ids starting with numbers to be set sometimes.

This commit changes the name for these drives to use sequential driveX numbering instead of UUID. Fixes #3388 

Tested successfully on macOS 12.1:

1. Place an image file in the Images folder of an existing UTM VM
2. Opening VM config
3. Click Drives tab
4. Change the new drive at the bottom of the list to type "disk image"
5. Click Save

Start VM → no error
Open Config QEMU tab → see drive id "driveX" instead of UUID.

### DIFF
--- a/Configuration/UTMQemuConfiguration+Drives.m
+++ b/Configuration/UTMQemuConfiguration+Drives.m
@@ -104,7 +104,7 @@ static const NSString *const kUTMConfigCdromKey = @"Cdrom";
 - (void)recoverOrphanedDrives {
     NSArray<NSString *> *orphans = self.orphanedDrives;
     for (NSInteger i = 0; i < orphans.count; i++) {
-        NSString *name = [NSUUID UUID].UUIDString;
+        NSString *name = [NSString stringWithFormat:@"drive%@", [[NSUUID UUID] UUIDString]];
         [self newDrive:name path:orphans[i] type:UTMDiskImageTypeNone interface:@""];
     }
 }


osy:
There's a subtle problem that occurs if the name is already used (say someone made 5 drives and deleted drive 4 then orphaned drive3 then we have two drive5)

Okay looking more into this I found something real dumb

So this commit fixed the issue https://github.com/utmapp/UTM/commit/b4b089070ec52a48766d600ebe611851ea3cb493

This commit literally reverted the change https://github.com/utmapp/UTM/commit/997be55e62baf88b721dcfa5d656bc8469462b74 because it introduced an error when drives are re-ordered.

I think a "proper" fix will have to involve finding an unused name but introducing FileManager into UTMConfiguration kinda breaks OOP principles.

Here's a proposal stupid enough for this stupid bug: make the orphaned drive name drive-UUID

PR #3410, opened 2021-12-25
Change default display upscaling mode to Nearest Neighbor

### LINKED ISSUE #3371: Change default display upscaling mode to nearest neighbor
It was suggested [on Twitter](https://twitter.com/rsms/status/1467604395302866948?s=20) to change the default upscaling mode to integer/nearest neighbor. This results in sharp (but blocky) graphics for low-res guests when the window resolution is greater than the host screen resolution.

I am in favor of this idea as the now-default linear upscale doesn't look pretty in my opinion. Here is the same VM in two screenshots. 

This screenshot taken with linear scaling mode:
utm-linear-scaling

This screenshot taken with nearest neighbor scaling mode:
utm-nearest-neighbor-scaling

Note the noticeable lower contrast in the menu bar text in the linear mode, as well as the general fuzziness.

conath:
Closes #3371
Nearest Neighbor offers better contrast and increases sharpness for upscaling;
the reverse is true for downscaling so it is not changed.

### DIFF
--- a/Configuration/UTMQemuConfiguration+Defaults.m
+++ b/Configuration/UTMQemuConfiguration+Defaults.m
@@ -42,7 +42,7 @@
         self.systemBootDevice = @"cd";
     }
     self.systemUUID = [[NSUUID UUID] UUIDString];
-    self.displayUpscaler = @"linear";
+    self.displayUpscaler = @"nearest";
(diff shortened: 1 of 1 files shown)


osy:
Can you rebase this to v3.0?

PR #3469, opened 2022-01-12
sharefilemodifier: take ShareItem as input, not function

ktprograms:
Forgot to do this in #3056


osy:
Can you explain why this change is needed?

ktprograms:
It makes the code cleaner by passing an instance of `ShareItem` instead of a function that creates a `ShareItem`

osy (approved):

PR #3484, opened 2022-01-14
macOS(qemu): add support for drives on external disks

ktprograms:
Full list of changes made:
- Add a `Reference` key to the drive configuration.
- Generate bookmark and access with it in argsForDrives if needed.
- Display drive on external disk with ` (reference)` appended.
- Add `reference` parameter to `importDrive`, used to change whether to
  perform a copy from source to dest, or to write bookmark data to dest.
- Add `newIndexCompletion` parameter to `importDrive`, used to return
  the new index after adding the drive. This is needed in
  `moveDriveToExternal` in order to maintain the boot order.
- Add `moveDriveToExternal` method that first performs the actual move
  of the drive, then calls `importDrive` with `reference` set to `true`
  in order to import and save a bookmark to the drive. It also then
  calls `moveDrive` in order to maintain the boot order.
- Change the icon used with the `Delete Drive` button to have an `xmark`
  instead of a `plus`, since that indicates removal/deletion.
- Add `Move to External Disk` button in the settings.
- Add `moveToExternal` function in `VMConfigDrivesButtons.swift` that
  opens a `NSSavePanel` with a default directory of `/Volumes` to ask
  the user where to move the drive to, then calls `moveDriveToExternal`
  with the selected destination.

Still to do:
- [ ] Error handling
- [ ] Possibly more when I think of them

### DIFF
--- a/Configuration/UTMQemuConfiguration+Drives.h
+++ b/Configuration/UTMQemuConfiguration+Drives.h
@@ -39,6 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)recoverOrphanedDrives;
 
 - (NSInteger)newDrive:(NSString *)name path:(NSString *)path type:(UTMDiskImageType)type interface:(NSString *)interface;
+- (NSInteger)newDrive:(NSString *)name path:(NSString *)path type:(UTMDiskImageType)type interface:(NSString *)interface reference:(BOOL)isReference;
 - (NSInteger)newRemovableDrive:(NSString *)name type:(UTMDiskImageType)type interface:(NSString *)interface;
 - (nullable NSString *)driveNameForIndex:(NSInteger)index;
 - (void)setDriveName:(NSString *)name forIndex:(NSInteger)index;
@@ -50,6 +51,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setDriveImageType:(UTMDiskImageType)type forIndex:(NSInteger)index;
 - (BOOL)driveRemovableForIndex:(NSInteger)index;
 - (void)setDriveRemovable:(BOOL)isRemovable forIndex:(NSInteger)index;
+- (BOOL)driveReferenceForIndex:(NSInteger)index;
+- (void)setDriveReference:(BOOL)isReference forIndex:(NSInteger)index;
 - (void)moveDriveIndex:(NSInteger)index to:(NSInteger)newIndex;
 - (void)removeDriveAtIndex:(NSInteger)index;
 - (NSString *)driveLabelForIndex:(NSInteger)index;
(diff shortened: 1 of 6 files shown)


ktprograms:
@osy Seeing that you plan to support multiple modes when importing (with the `copy:` parameter), and this PR adds a `reference:` parameter, would it just be better to use a `mode:` parameter with an `ImportModes` enum?

diepeterpan:
Thank you to everyone for adding this feature. I really hope this feature gets released soon, the current workaround https://github.com/utmapp/UTM/discussions/3355 is not easy.

ktprograms:
@osy I asked this in my previous PR on this topic, but I'll ask it here again:

What is the plan for an error message due to the bookmark not being resolvable? Is there any way to get the URL that the bookmark _would_ resolve to, and then show the Volume name that the URL was stored at?

nigamelastic:
I hope this feature gets shipped soon

osy (commented):
Thanks for these changes. While they work fine, I thought that having to maintain config.plist in a different directory than the data files is error prone, especially since we already have removable drives (now fixed drives can also be bookmarks). There's also an additional layer of complexity for users as they have to consider where individual drives reside (in .utm or externally). And also the issue of the file being missing or moved around (how do the user rectify the error?).

I've decided to take an alternative approach which is to allow importing `.utm` bundles as a shortcut as well as move `.utm` bundles elsewhere. Although not as flexible as moving individual drives (no support for multiple images in different drives for example), it is easier to manage for the user. All data is confined in the .utm bundle still and you can freely move and copy it around as needed. Files aren't scattered in different places and any "not found error" can easily be rectified as finding where the .utm bundle resides. This stays in tune with the philosophy of keeping all the data in a single bundle.

ktprograms:
> I've decided to take an alternative approach

@osy So are you implementing it?

Also will it work for Apple Virtualization?

osy:
https://github.com/utmapp/UTM/commit/8bad216f10464395db46ca048b5102c0c654c754

Apple VM doesn’t support saving the shortcut so you have to open the .utm every time you launch it.

ktprograms:
@osy Thanks for implementing it, seems to work fine.

Perhaps the Shortcut indicator arrow could be a bit bigger?

ktprograms:
I guess I'll close this PR then?

osy:
I don't see why it's not possible to keep files in sync if it's in a .utm bundle?

phirestalker:
> I don't see why it's not possible to keep files in sync if it's in a .utm bundle?

You might be right. I forgot that utm files are really just folders. On android where my app is I should be able to navigate into the utm folder and get directly at the container file, since the app does not read utm files.

osy:
Closing due to comment above.

PR #3502, opened 2022-01-18
config(apple): save config.plist as human-readable xml

ktprograms:
I have tested that this change is backward and forward compatible, meaning that I can save an xml config, and open it with UTM 3.0.3 (Beta) which expects a binary config (and if saving changes with 3.0.3 it will re-write it as binary), and I can also open a binary config with this code and when saving it will re-write it as xml.


osy:
I get this for encoding but why is it needed for decoding?

ktprograms:
Hmm, I thought specifying the format was needed for decoding, but it seems like it's detected automatically. That's probably why it's `inout`, because it just returns the detected format.

PR #3568, opened 2022-02-02
Add support for changing console colors, properly load available font list on macOS

ktprograms:
See the individual commit messages for more details, but here's an overview:
- Add `consoleTextColor` and `consoleBackgroundColor` values to `UTMConfigurable`, and implement the setters/getters in `UTMConfiguration+Display` that use a keyed archiver to convert the `NSColor` into `NSData` that can be stored in `config.plist`. (This is needed because if I just save the hex values, there is a lot of drift in the color picker because of the loss in converting to/from a hex string, and it's done on each color value change)
- Add `ColorPicker`s to the display console configuration.
- Add and call a Javascript method that sets the color values in the hterm prefs.
- Save the color settings as hex strings when using the apple virtualization backend. This is fine because the conversion is only done on load/save and not on every color change.
- Set the `consoleFontSize` to 12 by default if it's `nil` when using the apple backend.
- Change the `supportedConsoleFonts` method to list all the fonts in each family, to allow maximal customization.
- Implement `supportedConsoleFonts` for macOS (instead of just returning an empty array) to allow picking custom fonts.
- Set the font used in SwiftTerm based on the configuration for the apple backend.

<hr>

@osy Do you think I should squash the [`fix build on iOS`](https://github.com/utmapp/UTM/commit/111576cd781e3fe454d5f7e41de48260197b58a0) commit with the [first commit](https://github.com/utmapp/UTM/commit/18d9be4d702c89cb961fbe529f69428dee2d0078)? They're similar enough and the first one isn't really a 'working' commit because it can't build on iOS.

Also please take a look at the two `TODO`s (I'll post a comment on them since I want to ask a question about them), and remind me to remove them before you merge this.

<hr>

BTW, during my testing, more in the early stages of implementing this, I found that the VMs got reordered in the list, but I'm not sure if it might have been because it's debug builds etc. In the later parts of developing this PR I haven't found this problem any more, so I don't think it's an issue.

### DIFF
--- a/Configuration/UTMConfigurable.h
+++ b/Configuration/UTMConfigurable.h
@@ -15,7 +15,13 @@
 //
 
 #import <Foundation/Foundation.h>
+#if TARGET_OS_OSX
 #import <AppKit/AppKit.h>
+#define PlatformColor NSColor
+#else
+#import <UIKit/UIKit.h>
+#define PlatformColor UIColor
+#endif
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -31,8 +37,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, nullable, copy) NSString *consoleTheme;
 // TODO: Maybe use CGColor (But hard to handle)
-@property (nonatomic, nullable, copy) NSColor *consoleTextColor;
-@property (nonatomic, nullable, copy) NSColor *consoleBackgroundColor;
+@property (nonatomic, nullable, copy) PlatformColor *consoleTextColor;
+@property (nonatomic, nullable, copy) PlatformColor *consoleBackgroundColor;
 @property (nonatomic, nullable, copy) NSString *consoleFont;
 @property (nonatomic, nullable, copy) NSNumber *consoleFontSize;
 @property (nonatomic, assign) BOOL consoleCursorBlink;
--- a/Configuration/UTMConfigurable.h
+++ b/Configuration/UTMConfigurable.h
@@ -15,13 +15,6 @@
 //
 
 #import <Foundation/Foundation.h>
-#if TARGET_OS_OSX
-#import <AppKit/AppKit.h>
-#define PlatformColor NSColor
-#else
-#import <UIKit/UIKit.h>
-#define PlatformColor UIColor
-#endif
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -36,9 +29,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable, copy) NSString *notes;
 
 @property (nonatomic, nullable, copy) NSString *consoleTheme;
-// TODO: Maybe use CGColor (But hard to handle)
-@property (nonatomic, nullable, copy) PlatformColor *consoleTextColor;
-@property (nonatomic, nullable, copy) PlatformColor *consoleBackgroundColor;
+@property (nonatomic, nullable, copy) NSString *consoleTextColor;
+@property (nonatomic, nullable, copy) NSString *consoleBackgroundColor;
 @property (nonatomic, nullable, copy) NSString *consoleFont;
 @property (nonatomic, nullable, copy) NSNumber *consoleFontSize;
 @property (nonatomic, assign) BOOL consoleCursorBlink;
--- a/Configuration/UTMConfigurable.h
+++ b/Configuration/UTMConfigurable.h
@@ -15,6 +15,7 @@
 //
 
 #import <Foundation/Foundation.h>
+#import <AppKit/AppKit.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -29,6 +30,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable, copy) NSString *notes;
 
 @property (nonatomic, nullable, copy) NSString *consoleTheme;
+// TODO: Maybe use CGColor (But hard to handle)
+@property (nonatomic, nullable, copy) NSColor *consoleTextColor;
+@property (nonatomic, nullable, copy) NSColor *consoleBackgroundColor;
 @property (nonatomic, nullable, copy) NSString *consoleFont;
 @property (nonatomic, nullable, copy) NSNumber *consoleFontSize;
 @property (nonatomic, assign) BOOL consoleCursorBlink;
--- a/Platform/macOS/Display/VMDisplayTerminalWindowController.swift
+++ b/Platform/macOS/Display/VMDisplayTerminalWindowController.swift
@@ -85,11 +85,8 @@ extension VMDisplayTerminalWindowController: WKNavigationDelegate {
     
     func updateSettings() {
         if let consoleTextColor = vmQemuConfig?.consoleTextColor,
-           let consoleBackgroundColor = vmQemuConfig?.consoleBackgroundColor,
-           // TODO: Should use default colors, or keep it as if let?
-           let consoleTextColorString = consoleTextColor.hexString,
-           let consoleBackgroundColorString = consoleBackgroundColor.hexString {
-            webView.evaluateJavaScript("changeColor('\(consoleTextColorString)', '\(consoleBackgroundColorString)');") { (_, err) in
+           let consoleBackgroundColor = vmQemuConfig?.consoleBackgroundColor {
+            webView.evaluateJavaScript("changeColor('\(consoleTextColor)', '\(consoleBackgroundColor)');") { (_, err) in
                 if let error = err {
                     logger.error("changeColor error: \(error)")
                 }
--- a/Platform/macOS/Display/VMDisplayTerminalWindowController.swift
+++ b/Platform/macOS/Display/VMDisplayTerminalWindowController.swift
@@ -84,6 +84,17 @@ extension VMDisplayTerminalWindowController: WKNavigationDelegate {
     }
     
     func updateSettings() {
+        if let consoleTextColor = vmQemuConfig?.consoleTextColor,
+           let consoleBackgroundColor = vmQemuConfig?.consoleBackgroundColor,
+           // TODO: Should use default colors, or keep it as if let?
+           let consoleTextColorString = consoleTextColor.hexString,
+           let consoleBackgroundColorString = consoleBackgroundColor.hexString {
+            webView.evaluateJavaScript("changeColor('\(consoleTextColorString)', '\(consoleBackgroundColorString)');") { (_, err) in
+                if let error = err {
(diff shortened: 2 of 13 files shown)


Configuration/UTMConfigurable.h
```diff
@@ -28,6 +35,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable, copy) NSString *notes;
 
 @property (nonatomic, nullable, copy) NSString *consoleTheme;
+// TODO: Maybe use CGColor (But hard to handle)
+@property (nonatomic, nullable, copy) PlatformColor *consoleTextColor;
+@property (nonatomic, nullable, copy) PlatformColor *consoleBackgroundColor;
```
ktprograms, line 40: I was thinking maybe using a `CGColor` might be a good idea for two reasons:
- SwiftUI's `ColorPicker` can take a binding to a `CGColor`
- It would be compatible with both platforms, the `PlatformColor` stuff wouldn't be needed

But the problem is that `CGColor` isn't an object, so passing it around is harder. And also I'm not sure if `CGColor` can be used with a keyed archiver to save it.

I think it wouldn't be worth the extra trouble, since it isn't that much code in the `Binding` that converts a `NSColor` to the SwiftUI `Color` type.

Platform/macOS/Display/VMDisplayTerminalWindowController.swift
```diff
@@ -84,6 +84,17 @@ extension VMDisplayTerminalWindowController: WKNavigationDelegate {
     }
     
     func updateSettings() {
+        if let consoleTextColor = vmQemuConfig?.consoleTextColor,
+           let consoleBackgroundColor = vmQemuConfig?.consoleBackgroundColor,
+           // TODO: Should use default colors, or keep it as if let?
+           let consoleTextColorString = consoleTextColor.hexString,
+           let consoleBackgroundColorString = consoleBackgroundColor.hexString {
```
ktprograms, line 91: I don't think using a default value would be necessary here, since default colors are set when migrating the configuration, and also in the `Binding` in the configuration view.

Configuration/UTMConfigurable.h
osy: > But the problem is that CGColor isn't an object, so passing it around is harder.

Why is it any harder than say NSInteger?

ktprograms: @osy How would it be done? (What should the property attributes be and do I use `struct CGColor *` or `CGColorRef *`)?

osy:
> This is needed because if I just save the hex values, there is a lot of drift in the color picker because of the loss in converting to/from a hex string, and it's done on each color value change

Is that true though? I would think any drift would be unperceivable because colors are internally represented as 32 bit values (8 bit RGBA) and NSColor/UIColor decide to represent each component as a float. Even if it can physically represent more colors than RGBA8, it's pointless because your display won't be able to show it.

I would prefer to store the color values either as a hex string (without the #) or as a NSNumber (32 bit representation of the hex). First of all it makes modifying and viewing config.plist easier for debugging. Second, the size of config.plist smaller since we don't have to store all the extra info from the archived values.

Other than that, everything else looks fine.

Configuration/UTMConfigurable.h
osy: Look at my latest comment, I think NSNumber or NSString would be better fit here.

ktprograms:
@osy In my testing, the drift was so bad that you couldn't even select certain parts of the color wheel because it would drift to black when you went near them.

osy:
If you don't use `usingColorSpace(.sRGB)` does it make a difference?

ktprograms:
@osy that's for casting it to RGB so that the hex can be extracted.

Do you mean does the drift happen or what?

osy:
Okay I see that's right but then when you convert from hex back to NSColor you need to use https://developer.apple.com/documentation/appkit/nscolor/1532666-init because the hex is in sRGB.

ktprograms:
@osy Ok I'll change that.

Anything else that needs changing?

Configuration/UTMConfigurable.h
ktprograms: @osy The alternative solution would be to store an NSColor somewhere in UTMConfiguration, and then convert it to a hex string when saving the plist, like how it's done for the Apple backend.

If you want to do that, how/where should the intermediate NSColor be stored, and where can I put the conversion code?

osy:
No that’s it

ktprograms:
@osy Do you think I should squash the [`fix build on iOS`](https://github.com/utmapp/UTM/commit/111576cd781e3fe454d5f7e41de48260197b58a0) commit with the [first commit](https://github.com/utmapp/UTM/commit/18d9be4d702c89cb961fbe529f69428dee2d0078)? They're similar enough and the first one isn't really a 'working' commit because it can't build on iOS.

Configuration/UTMConfigurable.h
osy: I think it should be NSString in UTMConfigurable and both QEMU and Apple backends store the hex string.

Then when it's used for QEMU terminal, it uses the hex string directly. For Swift terminal it converts the hex string to NSColor (through your extension) in the view controller when it's being set.

ktprograms: @osy How do we prevent the drift then?

osy:
I think it's fine as is. We don't have any strict requirements on commits (yet). Some of these PRs have a really messy commit history already. Maybe in the future we'll set some better guidelines.

Configuration/UTMConfigurable.h
osy: Storage call .hexString

Usage call NSColor(fromHexString: String) which you implemented in the extension but you need to change the init there to call init(srgbRed:green:blue:alpha)

ktprograms: @osy You see, I can do that for the Apple backend because it's only converted when storing, do drift isn't a problem when selecting the color.

It's different with the Qemu backend because when setting or getting the console colors, it will have to do the conversion to save it into the dictionary. And because it's directly bound to the value used in the SwiftUI color picker, a small amount of drift on each conversion will be amplified to the point of not being usable.

osy: Can you try this different multiplier https://stackoverflow.com/a/39358741 seems like UIColor uses extended sRGB

ktprograms: @osy Nope, there's still a (slightly smaller but still noticeable/unusable) amount of drift.

I think the only solution if you want to store the console colors as a hex string is to only do the conversion when saving/loading the plist, not in the setter/getter for the color properties.

osy: Any chance you can ask stackoverflow with details on what you're seeing? I don't think there should be drift and maybe it's a bug that others have workarounds for.

ktprograms: @osy Done: https://stackoverflow.com/questions/71016043/swiftui-colorpicker-with-binding-to-hex-string-causes-large-drift

osy:
@ktprograms Can you give me write access to your fork so I can push changes? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

ktprograms:
@osy just did, can you please check if I did it correctly? Thanks!

osy:
I've rebased and add my proposal to store the hex string and compute NSColor/Color on demand. I was not able to observe the colour shift problem on either macOS or iOS simulator. Can you test it and see? If you still observe the shift, can you please provide a screen recording so I can better see what the issue is?

ktprograms:
@osy I'm sorry, but I'm away from my computer for the next two days, so I can't test it.

osy:
https://user-images.githubusercontent.com/50960678/155071330-8bd58527-66e0-42a3-965a-8205e264bd3a.mp4

Okay can you look at the screen recording and see if you see any issues? I'm unable to reproduce this "shift" issue.

ktprograms:
@osy unfortunately,  I don't have a fast internet connection currently, so can you please compress this video as much as possible? Thanks!

Did you also test it on macOS? It's possible that it's a Big Sur bug because that's what I used to test.

osy:
https://user-images.githubusercontent.com/50960678/155077201-cd82962d-ea6c-415b-8c4c-17a7c7453a53.mp4

I don't have Big Sur installed currently, but I think I'm fine with merging it as-is and the next beta if people have issues we can fix it afterwards.

ktprograms:
Ok, thanks for merging!

ktprograms:
@osy The drift is still there, but it's _usable_.

For example, in the iOS simulator, when using the Sliders tab of the color picker, you can see in the video below that when dragging one slider, the other two move slightly (look at the numbers). Note that this is also visible in your original screen recording as well.

https://user-images.githubusercontent.com/71804605/155639806-7fbef4c8-2346-464f-905d-37066b981396.mp4

On Mac, the drift is also there, you can see the brightness slider moving when going around the color wheel, which doesn't normally happen if not converting to a string.

https://user-images.githubusercontent.com/71804605/155640269-3cd8ef73-b5f0-40ea-b02a-6258546bb0bf.mp4

osy:
I see what you mean. I guess it wasn't noticeable just by looking at the colour. My best guess is that SwiftUI's Color class isn't respecting SRGB vs P3. I will file feedback with Apple about it.

osy:
So to expand on why you're seeing the drift: P3 colour space is larger than sRGB (it can represent more colours). When you convert from P3 to sRGB, it will truncate any value that it cannot represent into the closest one where it can represent.

iOS's ColorPicker has the option to switch between P3 and sRGB. If you choose sRGB, then it will not try to set any value that it cannot represent (no conversion is done, the picker just skips over the values that do not exist). The space is large enough that it's not perceivable to the user.

However, a bug in SwiftUI makes the picker automatically switch to P3 when you pass in a newly created Color() object (even if you specify sRGB in the init). That means as soon as you change the colour slider, it switches to the P3 selector, but then it tries to store the P3 value and gets truncated which gets converted to sRGB and then back to P3 in ColorPicker which you see as a jump in the slider.

In macOS, it seems to always be in P3 mode and there's no way to switch it in the UI.

So at the end of the day here's the fix/workaround: ff79144e63c0cfb915d7b0fce0bb875c4ab2e738

1. We always save the P3 hex, NOT the sRGB hex
2. This is fine for SwiftTerm which doesn't use the hex at all
3. In hterm.js we have to convert from P3 to sRGB which is a lot of code but we will soon be removing hterm.js so that code will be deleted along.

ktprograms:
@osy Thanks for looking into it! I can confirm that there is no drift on Mac.

PR #3622, opened 2022-02-13
Add Pop!_OS icon

conath:
This adds the icon for the Ubuntu-based OS from system76, [Pop!_OS](https://pop.system76.com).

Here is how it looks in UTM, light and dark mode:
utm-pop-os-light
utm-pop-os-dark


conath:
Cancelled CI because it's just a icon added, which is not covered by CI anyway.

PR #3655, opened 2022-02-21
Added full screen window resolutions for MacBook Pro 2021

pointum:
Full screen window size of a macOS VM does not match the native resolution of the screen if there's a notch.
I've added resolutions that are 74 pixels less in height, so that the guest OS resolution matches without scaling.

UTM Full Screen

UTM Display Settings

Also updated the list with the × sign to match macOS style:

System Preferences Displays


osy:
Hmm maybe I should label the other entries as well...

osy (approved):

PR #3700, opened 2022-02-28
UI polish for macOS

j-f1:
I did a pass over most of the configuration UI and fixed many of the issues I saw. See commit messages for full details on what’s changed.

Let me know if I should break this down into multiple PRs or add/remove any changes!

Some highlights:

Screenshot_2022-02-28 00 43 19
Screenshot_2022-02-28 00 42 36
Screenshot_2022-02-28 00 42 42


j-f1:
There are still a few things that I don’t really like, but hopefully future versions of SwiftUI will have nicer support for `Form` on macOS so it gets easier to make great UIs.

osy:
Hi thanks for this. Before I review the PR, I have a few requests
1. Can you rebase it on the most recent commit as I've done quite a bit of tweaking of the UI myself.
2. Can you make sure to test on the following: macOS 11.x, macOS 12.x, iOS (Simulator) 14.x, iOS (Simulator) 15.x. From experience, SwiftUI appears quite differently between OS updates.

Thanks!

j-f1:
Sure! I’ll let you know when I do that. Do you have any tips for testing on macOS 11? I only have access to a macOS 12 machine locally.

osy:
If you have an Intel mac, you can use Parallels or VMware to install Big Sur. Otherwise, leave it out and I'll test it.

osy:
Heads up, I made more changes to the layout but things should be more stable after today.

j-f1:
Tested on:
- macOS 12.2.1 (21D62) (Apple Silicon)
- iPhone 13 (simulated, iOS 15.2)
- iPhone 12 (simulated, iOS 14.5)

osy (changes requested):
In HEAD, we removed the advanced section all-together and reformatted RAMSlider to fix the layout weirdness. With that gone, I think the advanced tab for network settings should also be moved back so there's no single advanced sticking out. I'm also debating on removing the advanced toggle from network as well.

EDIT: Sorry I was reviewing this in the order of the commit and I see the revert later. Seems like Github doesn't attach the commit that I was looking at when making the comment...

osy:
I think the changes to move IP Configuration to its own page can be a separate PR. I'm trying to decide how to keep a consistent UI on iOS and macOS while on iOS you normally don't see an indented entry.

Also the change to use a "Browse..." button (the old design) on the details view should be a separate PR so we can discuss it some more.

Then if you can fix up the macOS CI failure and make sure I have write permission on your fork, I can try to merge the changes.

Thanks!

osy (approved):
I've played around with this a bit more in iOS and macOS and I think it's okay as is. I see that the network settings is different on iOS vs macOS and the removable/shared picker is fine as well.

osy:
I've also tested it on Big Sur and aside from the macOS fixes I posted, it works fine as well.

j-f1:
Thanks! Sorry I wasn’t able to get to your suggestions, things have been a bit busy :)

ktprograms:
@osy @j-f1 The notes field in settings no longer has a border on macOS (was introduced in https://github.com/utmapp/UTM/commit/46722786d195b2586e4a796f95eec9610e920629), because the refactor in https://github.com/utmapp/UTM/commit/7c0fbe8f45be70d3383f981d34c618a781205e20 removed it.

It _should_ be as simple as changing
https://github.com/utmapp/UTM/blob/17e18d5f73b0b569a4e8e63a8f6dc250038ac7c5/Platform/Shared/VMConfigInfoView.swift#L84-L87
to be
```swift
    private var notesField: some View {
         TextEditor(text: $config.notes.bound)
              #if os(macOS)
              .border(Color.primary, width: 0.5)
              #endif
             .frame(minHeight: 200)
     }
```

osy:
Thanks I fixed it in https://github.com/utmapp/UTM/commit/1856e4076b8abfe8d6a5d015b8feaaed80f3107d

Issue #3913, opened 2022-04-19
Reduce dynamic library binary size


Halo-Michael:
Some existing dynamic library files have useless slices, please process them to reduce installation package size
![EC68BFE5-B61A-460B-A375-F4FA543098F7](https://user-images.githubusercontent.com/45062120/163924758-9fe5adad-d4cd-4f22-a213-cd77da8d6724.png)
.

Halo-Michael:
The processed installation package can reduce the size of about 36mb.
![466B059E-C09E-4FBB-AF71-56E597C421C0](https://user-images.githubusercontent.com/45062120/163926473-75e50577-a32d-47e2-9990-f2bd6252279f.png)

PR #3923, opened 2022-04-22
Fix issue #3913

Halo-Michael:
Little update, with https://github.com/utmapp/UTM/issues/3913 fix.


Halo-Michael:
Due to lipo will print error when binary is not a fat binary file, so we need to check it first, turn to https://github.com/utmapp/UTM/pull/3934.

PR #3934, opened 2022-04-23
Fix issue #3913

Halo-Michael:
Fix https://github.com/utmapp/UTM/issues/3913


Halo-Michael:
Well, why binary build for arm64, ios_simulator does not have arm64 slice? @osy 
CC: https://github.com/utmapp/UTM/runs/6138999484?check_suite_focus=true#step:9:3068

osy:
I'm not sure, it works fine though so maybe check the file and see what it contains?

Halo-Michael:
> I'm not sure, it works fine though so maybe check the file and see what it contains?

UTM main binary have arm64 silce, but the framework does not.
~K3(3V(I~MVIWR 42$W}4N3

Halo-Michael:
I guess we can only lipo for non-simulator build.

Halo-Michael:
Okey, that's it, the script will cut useless architectures for macOS/iOS now.

osy:
I'm pretty sure this breaks macOS universal which has both x86_64 and arm64.

Also why are you looking at `*/Frameworks/*.framework` I don't think there's any fat binaries there?

Overall I see the space reduced by 10% to 746MB for iOS. Is that really that big of a deal?

Halo-Michael:
> I'm pretty sure this breaks macOS universal which has both x86_64 and arm64.
> 
> Also why are you looking at `*/Frameworks/*.framework` I don't think there's any fat binaries there?
> 
> Overall I see the space reduced by 10% to 746MB for iOS. Is that really that big of a deal?

That sounds weried but only remove arm64e silce is not enough change a fat file to non-fat file, it's still a "fat file" even it may only have one silce:
34ACE7E2-3C14-404F-B96A-0B908F34F057
I guess we can only cut useless for iOS?

osy:
So is there any reason to include .framework?

Halo-Michael:
> So is there any reason to include .framework?

E80FCE81-25B7-46D1-AC6A-DE0510787A5D

Halo-Michael:
> > So is there any reason to include .framework?
> 
> E80FCE81-25B7-46D1-AC6A-DE0510787A5D

Fat binary probably takes up a full 16K as a page table or something else, so why not?

conath:
Thanks @Halo-Michael file size is important. 👏

Issue #3936, opened 2022-04-23
Delete existing screenshot when option is enabled


VariableDeclared:
Hi,

I had a quick search of open issues and could not see one that matched up; so here's the feature request.

When using UTM; UTM is periodically taking screenshots of the VM display, and, if the VM is shut down, that preview will remain. This is a privacy concern for users of UTM if the VM manager is effectively displaying their last activity, including open browsers, names, etc clearly displayed.

This request is to add an option disabling VM preview in the VM manager screen.

Many thanks,
VD

VariableDeclared:
Okay, so I did some digging, there's an option "Do not save VM screenshot to disk".

<img width="516" alt="Screenshot 2022-04-23 at 19 07 47" src="https://user-images.githubusercontent.com/7092570/164934358-10b507ad-eaa1-4add-9060-2c02dd77fe48.png">

Enabling this will achieve the desired result **but** it will not remove screenshots that already exist. So I found a workaround -:

1. Open your favourite terminal
2. `USER=$USER_MACOS_USERNAME`
3. `cd /Users/peterdesousa/Library/Containers/$USER/Data/Documents`
4. `ls` and note you should see your VMs in this folder. Ending `.utm`. These are folders, cd into your target VM
5. Once in the VM you can run `rm screenshot.png`
6. Exit UTM
7. Reopen - the screenshot was removed for me.

I think I can close this issue for now; given that I have provided a workaround for screenshot purging,

Cheers,
VB

osy:
Would it be useful if we delete the last saved screenshot the next time a VM is launched after “do not save screenshots” is checked?

VariableDeclared:
hey @osy yes, I think after enabling the feature purging should occur,

VD

VariableDeclared:
hey @osy I created a draft which might be suitable for addressing this issue, let me know what you think,

PR #3937, opened 2022-04-24
Vd/disable screenshots

VariableDeclared:
Hi, 

When disabling screenshots on UTM prefences the user cannot delete their screenshots after selecting "Do not save screenshots". This merge proposes a new configuration value "Enable screenshots" which will enable/disable them as well as delete them on disk when disabled.

Addresses https://github.com/utmapp/UTM/issues/3936

Cheers

Peter

### DIFF
--- a/Configuration/UTMAppleConfiguration.swift
+++ b/Configuration/UTMAppleConfiguration.swift
@@ -57,6 +57,8 @@ final class UTMAppleConfiguration: UTMConfigurable, Codable, ObservableObject {
     
     @Published var consoleResizeCommand: String?
     
+    @Published var isScreenshotEnabled: Bool
+    
     var iconUrl: URL? {
         if self.iconCustom {
             if let current = self.selectedCustomIconPath {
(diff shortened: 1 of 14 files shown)


osy:
Thanks for the PR but I think this overly complicates things. Also, an accidental check of the option shouldn't go and delete (screenshot) files from every VM without an additional prompt or something. I think it's enough to just delete any stale screenshot when the VM starts and the disable option is checked.

e958cc42c4eec2d0cf3e3bf0d3eb9403edd0b79b

VariableDeclared:
Cool thnx @osy

adevart:
Is there any possibility for an option to disable screenshots entirely or allow manually setting an image?

It's a security risk having screen capturing running constantly in the background. Even with the option disabled, it saves every 30 seconds so can still screenshot sensitive info like login details or emails and this appears on closing the VM, which then gets flushed on closing the app.

I'd like to be able to disable the screen capturing process and also set a VM screenshot manually and have it persistent so that different Windows and Linux VMs have different screenshots each.

PR #4067, opened 2022-05-26
I don't know if this is intentional...

Halo-Michael:


conath (approved):

osy:
No longer needed in the reworked config backend. Thanks.

PR #4087, opened 2022-06-02
Support drive resizing

stewartsmith:
Add UTMLogToMemory to capture output of "qemu-img info", which we can parse the JSON output of in order to determine the virtual size of the drive.

Use the same widget style for resizing as we do for creating a drive (GB/MB), although require clicking "Resize" to do the resize itself.

Currently doesn't warn about having not clicked "resize" though.

Fair warning: I'm pretty sure this is the first time in nearly 20 years I've looked at Objective-C code, and my Swift knowledge is sketchy at best.

Screen Shot 2022-06-01 at 22 58 28

### DIFF
--- a/Managers/UTMLogToMemory.swift
+++ b/Managers/UTMLogToMemory.swift
@@ -1,0 +1,32 @@
+//
+// Copyright © 2022 Stewart Smith. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


osy:
Thanks for the PR. There has been a lot of changes to the backend, so I picked up your changes to UTMQemuImage (using the new delegate interface for UTMLogging) and then reworked the UI to be coherent across all the different disk operations.

PR #4110, opened 2022-06-08
Fix MacOS build instructions

sheerun:
It seems that homebrew now installs to `/opt/homebrew`


osy:
The build scripts and instructions have been changed since v4 and this PR no longer applies.

Issue #4141, opened 2022-06-18
macOS VM Installation: Automatic .IPSW file download fails at 100%


LostInCompilation:
**Describe the issue**
I created a new VM by clicking "Virtualize->macOS" and provided no .IPSW file for import to UTM. After the initial VM creation, the download of the IPSW file starts and runs smoothly. However as soon as the download finished, I get the error message shown in the screenshot. After clicking "OK" no VM is created.

UTM seems to successfully create a .tmp file in "/private/var/folder/.../com.UTMApp.app/.../longfilename.tmp" during download. But this file is removed right after the error message pops up. I searched my file system, no other .tmp file or .IPSW file was present.

**Configuration**
* UTM Version: 3.2.4 (58)
* OS Version: macOS 12.4
* Device: M1 Pro MacBook Pro 16"

**Error message**
<img width="338" alt="Screenshot 2022-06-18 um 04 40 11" src="https://user-images.githubusercontent.com/12819635/174422541-23033a4c-b0d4-4bb6-b50f-2aef5b98d2f0.png">

conath:
The file path during the download is controlled by macOS, not UTM. Once it finishes, UTM receives the file path and tries to move it to its own temporary folder. Here is the code that does this:

https://github.com/utmapp/UTM/blob/753564cd3abc9db25fd94266781010bab6d9b5ab/Platform/Shared/UTMDownloadIPSWTask.swift#L36-L42

It seems like it does not check if the [caches folder](https://github.com/utmapp/UTM/blob/753564cd3abc9db25fd94266781010bab6d9b5ab/Platform/Shared/UTMDownloadIPSWTask.swift#L27) exists, and therefore might fail. A fix could be to check if the folder exists and create it if not.

osy:
Fixed by #4206

PR #4206, opened 2022-07-11
When downloading macOS IPSW: Check if "Caches" folder exists and create it if not

LostInCompilation:
Fix for #4141 
Just checking if the folder exists and create it if not. The problem with #4141 was a missing "Caches" folder.


conath (commented):
Looks good to me, except I am wondering if “create intermediate directories” should be ‘true’? Pinging @osy to approve & merge.

LostInCompilation:
I've set it to false to be a bit conservative. The path is this: /Users/NAME/Library/Containers/com.utmapp.UTM/Data/Library/Caches

com.utmapp.UTM/Data/Library should always be there.

PR #4529, opened 2022-10-20
Fix URL for Windows installation guide

lassebm:
The current URL is 404.


osy:
Thank you, this was fixed with another incorrect link as well. We also updated the documentation site to redirect on the broken URL.

PR #4592, opened 2022-11-01
Add Polish localization

mavethee:


osy:
Note to self: need to add to the project file. Can do it during merge.

PR #4610, opened 2022-11-04
Fix Settings UI on large content

JezerM:
This PR fixes some issues related to the Settings UI, where large content may be outside of the main content and therefore not displayed.

Before this PR, the content is not resized to the parent:

Captura de pantalla 2022-11-03 a la(s) 19 53 21

With this PR, the content is resized to fit the parent:

Captura de pantalla 2022-11-03 a la(s) 19 56 26


osy:
Need to test this on each macOS version due to SwiftUI differences

- [ ] macOS 11
- [ ] macOS 12
- [ ] macOS 13

PR #4805, opened 2022-12-15
Support Apple Events and introduce CLI tool to interface with it

### LINKED ISSUE #4331: Add shell commands to start a VM or check it is running
I'd really like a way to start a vm from the shell or via apple script as well as commands that allow me to check a vm is running.

I'm using a utm vm to sandbox a Cisco AnyConnect VPN that I really don't want to install / execute on my machine, because it forces all traffic to go through the vpn instead of allowing me a split tunnel setup.

What I'd like to do is have a shell script that starts the vm, waits for it to start successfully, then connects to it via ssh, starts the vpn.

Everything besides starting the vm if it is not running is already working - so there is one last bit to automate. Are you guys interested in supporting such use cases?

I would guess that a good way to do this would be to add command line argument support to the UTM binary, but of course many ways are possible.

osy:
We initially built a whole automation backend based off of a Swift<->JSON<->Unix Sockets protocol. However, last minute before release, we decided to scrap the whole thing and instead use OSA because it is more "Mac friendly" and is also better integrated with Apple scripting.

The initial set of commands allow limited querying of VM information (mostly just the items you would see in the GUI) as well as basic start/stop/suspend operations as well as a way to get the serial port address. Future commands and capabilities can be added as needed and should be relatively straightforward.

Closes #4331


saagarjha:
"we"

PR #4866, opened 2022-12-31
Add TSO support for jailbroken/Trollstore iOS Hypervisor

osy:
This allows for faster x86 emulation inside the arm64 guest. Both Rosetta (patched for QEMU) and FEX-Emu can take advantage of this.


gabefair:
I updated the rosetta documentation to add instructions on confirming Rosetta is working inside LInux. Which would help newbies a lot.

https://github.com/utmapp/docs.getutm.app/pull/76

PR #4881, opened 2023-01-03
Fix typo + Update Japanese localization

MMP0:


njlyf2011:
How can I update the strings in the translation file to the latest version?
I want to update the strings in the Chinese translation, but I don't know how to do it.

PR #4918, opened 2023-01-07
Removed unspported OS logos

LostInCompilation:
Removed OS logos, which can definitely not run on UTM/qemu.

Removed:
* Apple TV OS
* iOS
* maemo
* Nintendo Switch OS
* PlayStation
* PlayStation Portable


osy:
Why though? Other than you nobody has ever complained about this. Why can’t I use the psp logo for my VM?

LostInCompilation:
It's not a complaint. There is just no real use for them or could lead to false expectations and questions on how to run them. It's just a cleanup. I mean you also wouldn't add random logos like from car brands.
But it's your choice, I just see no reason in keeping them.

osy:
I think unless we get a good reason (like a copyright claim), we don’t need to touch it.

LostInCompilation:
Well if "not able to run these systems at all" isn't a good reason then okay 😁

osy:
I mean you should also remove android then because even if you can boot a very old version it’s unusable. Maybe also remove windows 11 on iOS SE version? Etc. there’s a lot of ones that you “shouldn’t” pick.

LostInCompilation:
That is a bit nitpicky. Yes there may be OSes which wont run good or only run on Mac. But this is not comparable to a PlayStation which can not be emulated and doesn't even has an OS which could be installed.

osy:
Here’s a valid use case: say I want to run PCSX in a VM and that’s the only use for my VM. Well I would want to name it “PSX Emulator” and give it a PS1 logo.

LostInCompilation:
It's fine 😅 You don't need to defend your decision. It was just a first attempt to clean up the logos. If you don't want it, that's okay.

Issue #4940, opened 2023-01-13
OptionsList for Flags doesn't show pretty values, but rather raw values


stonerl:
I was toying around with UTM and wanted to change some prettyValue in the Flag list. But I noticed that these are never used and the raw values are shown instead.

https://github.com/utmapp/UTM/blob/46f1f22b22a45cc73d0377f57ea812fcf8ac9b02/Platform/Shared/VMConfigSystemView.swift#L217-L235

E.g. change line 4938 from `3dnow` to `3dyesterday`

https://github.com/utmapp/UTM/blob/46f1f22b22a45cc73d0377f57ea812fcf8ac9b02/Configuration/QEMUConstantGenerated.swift#L4936-L4938

It should now show `3dyesterday` in the UI but it uses the raw value from line 4616 instead.

https://github.com/utmapp/UTM/blob/46f1f22b22a45cc73d0377f57ea812fcf8ac9b02/Configuration/QEMUConstantGenerated.swift#L4615-L4616

Version/Platform: git-master on macOS

osy:
While you are technically right, there is not a single flag which doesn't have the same text for rawValue and prettyValue.

Issue #4914, opened 2023-01-07
Improve speed for x86 Windows guest by enabling Hyper-V Enlightenments by default


stonerl:
Currently, with the default settings for Windows guests on x86 hardware, the guests are very sluggish. To improve speed and the overall responsiveness, the following settings should be applied by default.

1. Disable **HPET** support `-no-hpet`[1]
2. Set the following **Hyper-V Enlightenments** [2] by default: `hv-relaxed,hv-vapic,hv_spinlocks=0x1fff,hv-runtime,hv-time,hv-frequencies,hv-xmm-input`

This increases speed (~20 % in single core & ~11 % in multicore) and responsiveness of the VM dramatically. I tested this with Geekbench 5 (Windows 11 guest):

[Bare Metal](https://browser.geekbench.com/v5/cpu/19750858): 1241/4766
[Default settings](https://browser.geekbench.com/v5/cpu/19751026): 848/3022
[With Hyper-V Enlightenments](https://browser.geekbench.com/v5/cpu/19750790): 1017/3375

[1] [i386 target only](https://www.qemu.org/docs/master/system/invocation.html?highlight=-no-hpet#hxtool-4)
[2] [Hyper-V Enlightenments](https://www.qemu.org/docs/master/system/i386/hyperv.html#hyper-v-enlightenments)

osy:
This works only on Intel with HVF I presume?

stonerl:
> This works only on Intel with HVF I presume?

I guess (I used it with HVF on Intel) , from what I read in the documentation. If someone with an Apple Silicon Mac could test whether these Enlightenments are also exposed there, for x86 guests, would be appreciated. Since I don't own such a device I can't test this myself.

What is the standard Hypervisor an Apple Silicon for x86? TCG or HVF?

osy:
HVF uses hypervisor. TCG is an emulator and recompiles the code so I doubt it works here.

stonerl:
Sorry, I expressed myself incorrectly. I'm aware that TCG doesn't use the Hypervisor. What I meant to say was, whether HVF can be used on Apple Silicon for x86 guests or if TCG has to be used.

But after reading the QEMU documentation again it seems that the Enlightenments are only available on x86 hosts for x86 guests.

osy:
[...]

PR #4951, opened 2023-01-17
Added Hyper-V Enlightenments for x86_64 and i386

stonerl:
I added some Hyper-V Enlightenments [1] to the UI. Non of them are active by default.

Enlightenments that require 'hv-vpindex' are omitted, since this enlightenment is currently not available on macOS.

Also, Enlightenments that are used for nesting e.g. 'hv-evmcs', or only change the Hyper-V Version ID 'hv_version-id-*' are omitted, as well.

Enlightenment 'hv-spinlocks' requires a parameter, currently 0x1fff is hardcoded since this seems to be a sane value. [2]

This commit partially addresses #4914 but also suffers from issue #4940.

Currently this only works for newly created VMs.

[1] https://www.qemu.org/docs/master/system/i386/hyperv.html
[2] https://leduccc.medium.com/improving-the-performance-of-a-windows-10-guest-on-qemu-a5b3f54d9cf5#c5b9

### DIFF
--- a/Configuration/QEMUConstant.swift
+++ b/Configuration/QEMUConstant.swift
@@ -124,6 +124,10 @@ protocol QEMUCPUFlag: QEMUConstant {}
 
 extension AnyQEMUConstant: QEMUCPUFlag {}
 
+protocol QEMUCPUHyperV: QEMUConstant {}
+
+extension AnyQEMUConstant: QEMUCPUHyperV {}
+
 protocol QEMUDisplayDevice: QEMUConstant {}
 
 extension AnyQEMUConstant: QEMUDisplayDevice {}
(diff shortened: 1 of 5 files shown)


osy (changes requested):
I made some source level comments but I have some higher level design concerns that supersedes that (i.e. if addressed, the source level comments may not be applicable).

There has never been a design goal of UTM to expose all QEMU options. Instead, we only expose options when there is a clear benefit to 90% of the users (as a rule of thumb). It seems from your comment in #4914 that there is a specific set of hyper-v entitlements that will speed up windows guests. In that case, my proposal is a checkbox in QEMU Tweaks for "Enable Hyper-V Enlightenments" which we default to on whenever the user selects Virtualization and Windows in the wizard and is on Intel. If that box is checked, we add the recommended string to the QEMU arguments. I don't see much value in exposing the complexity of all the different enlightenments to the user.

PR #5198, opened 2023-04-16
Add AlmaLinux icon

andrewlukoshko:
AlmaLinux is a "classic" (or pre-stream) CentOS successor.
https://almalinux.org


andrewlukoshko:
@osy could you check this please? Thanks.

PR #5387, opened 2023-06-11
Update Japanese localization

studio1a9:
Since "Dock" is a common translation for Apple, I propose to write it in English instead of Japanese.
We also propose Japanese translations for messages that are not localized in Japanese.


MMP0:
Hi, you’re totally right. I always refer to the translations by Apple but I completely missed “Dock”.
Thank you for finding this!

osy:
Closing since #5357 has already merged the changes.

PR #5412, opened 2023-06-29
Update TetheredLaunch.md

startergo:
Apps with the Debugging Tool Entitlement can call task_for_pid() to retrieve a valid task port for unsigned and third-party apps with the Get Task Allow entitlement set to true. However, even with the debugging tool entitlement, a debugger can’t get the task ports of processes that don’t have the Get Task Allow entitlement, and that are therefore protected by System Integrity Protection.


saagarjha:
System Integrity Protection is not relevant on iOS

startergo:
> System Integrity Protection is not relevant on iOS

https://github.com/osy/Jitterbug

```
Launch fails with "Failed to get the task for process xxx"

This application does not have the get-task-allow entitlement. Or this is not a development app.
```
Jitterbug will output an error if ```No get-task-allow``` is checked.

PR #5432, opened 2023-07-07
Fix call of usage when packing dependencies for one architecture only

gahfy:
The doc stated to run `./scripts/pack_dependencies.sh . macos arm64 x86_64` to pack dependencies for universal binaries.

However, if you just want to build it for running it locally with the command `./scripts/pack_dependencies.sh . macOS arm64`, the `usage()` method will get called.

This fix let you pack dependencies for one architecture only.

### DIFF
--- a/scripts/pack_dependencies.sh
+++ b/scripts/pack_dependencies.sh
@@ -11,7 +11,7 @@ usage () {
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


osy:
Sorry, what's the purpose of running this with one architecture?

PR #5550, opened 2023-08-10
Improve Simplified Chinese Translation

changanmoon:
Along with a small change in `Platform/macOS/Display/VMDisplayQemuDisplayController.swift`

(Note that two strings have been changed in this file, the localized ones must be changed accordingly. I only made modifications to Simplified Chinese)


osy:
FYI: When you change the base string (the “…”), it breaks all the other translations.

changanmoon:
> FYI: When you change the base string (the “…”), it breaks all the other translations.

Okay, I took back that change :)

changanmoon:
Finally succeeded! Sorry for the inconvenience…
Additionally I updated Simplified Chinese translation of README, hope this helps :)

PR #5680, opened 2023-09-07
Rewrite Services & iOS Display Logic in Swift

IsaacMarovitz:
- [x] UTMProcess
- [x] UTMQemuSystem
- [x] UTMLocationManager
- [x] UTMLogging
- [x] UTMSpiceIO
- [x] Fix SPICE connections being refused
- [x] Change `startProcess` to use scoped UnsafePointer to resolve dangling pointer warning
- [x] VMScroll
- [x] VMCursor
- [x] VMDisplayMetalViewController
- [x] VMDisplayMetalViewController+Pointer
- [ ] VMKeyboardButton
- [ ] VMKeyboardView
- [ ] VMDisplayViewController
- [ ] VMDisplayMetalViewController+Touch
- [ ] VMDisplayMetalViewController+Pencil
- [ ] VMDisplayMetalViewController+Keyboard
- [ ] VMDisplayMetalViewController+Gamepad


osy:
Just FYI, we purposely left objective c code in objective c if it’s more reasonable to use it due to either C library or Apple framework compatibility or because the code is cleaner.

IsaacMarovitz:
> Just FYI, we purposely left objective c code in objective c if it’s more reasonable to use it due to either C library or Apple framework compatibility or because the code is cleaner.

I agree that there are some instances here where the Swift solution is significantly messier case in point, `startProcess` on `UTMProcess`. However, there *are* things here that can be converted to Swift with a clean solution. I'm happy to revert some of the rewrites if you decide they're too messy for your taste.

How it's written now is very much the "cheap-n-dirty" solution and not necessarily the cleanest "Swiftiest" way. I will work on improving the messier parts.

osy:
My philosophy usually is I don’t rewrite any code for the sake of rewriting it. It’s only if I need to implement a feature or I want to rearchitect a particular part or something. Case in point, the last rewrite I refactored a lot of the QEMU support code into QEMUKit and made swift friendly interfaces. I think that out of everything, the backend stuff least need to be converted to swift. There’s a lot of front end iOS stuff that’s still objc from the early days and moving to swift may actually benefit the code.

Anyways feel free to do what you want, just letting you know my thoughts early so you don’t waste any time if it doesn’t get merged.

PR #5756, opened 2023-09-30
Update iOSDevelopment.md

Vladlen45:


osy:
I'm not sure this is correct. You should use the `./scripts/package.sh` which will do the signing. This is detailed under `Packaging`

Issue #5898, opened 2023-11-14
Main window not listed in window menu


rbgarga:
Main UTM window, the one that lists all VMs and shows a preview of VM console is not listed in Window menu.  If it's accidentally closed there is no way to show it again except if you quit the app and start it over.

I'm running version 4.4.4 (92) installed using Homebrew on macOS Sonoma 14.1.1

osy:
What happens if you click the dock icon?

rbgarga:
> What happens if you click the dock icon?

It shows the window of running VMs

rbgarga:
If I press `cmd+n`, to create a new VM, the main window does show up

MitchTalmadge:
This is still an issue. If I close the main window and a VM is running, there's no obvious way to open back up the main window. If I quit UTM and then re-open it, then I can get the main window.. but that shuts down the VMs, obviously.

Looks like this is similar/the same as #3632

PR #5908, opened 2023-11-18
macos: add Window>Show UTM menu option

garamb1:
As mentioned in issue https://github.com/utmapp/UTM/issues/5898, I also noticed there is no way of getting only the virtual machine list window to open if:
 - UTM is running one of more virtual machines
 - The virtual machine list is closed
 - The menu bar icon & menu are disabled (hence the user cannot use CMD + 0 or click on `Show UTM` from the menu)

This PR adds the option in the Window Menu to show UTM, also adding the CMD + Alt + L keyboard shortcut to do so.

The main window is shown by sending a new notification, handled in the macOS `homeWindow` View. I am not sure whether this would be the best way of doing so with SwiftUI, so any suggestions are welcome! Thanks for the amazing app!


garamb1:
@osy thanks for adding this to the next release!
I had to cherry pick my original commit and merge the changes from main, it should be ok now.

Platform/macOS/UTMApp.swift
```diff
@@ -32,6 +32,9 @@ struct UTMApp: App {
                     data.showErrorAlert(message: message)
                 }
             }
+            .onReceive(NSNotification.ShowMainWindow) { _ in
+                NSApplication.shared.windows.first?.makeKeyAndOrderFront(nil)
```
osy, line 36: I don't know if this is good enough. I might play around with it but thinking of capturing a reference in the AppDelegate in order to reference it.

garamb1: Sounds good, thanks!

osy:
I just realized there is already an option to show the main window:
image

Also, I don't understand this:
> The menu bar icon & menu are disabled

image

When you disable the dock icon, the menu icon is forced to be enabled.

garamb1:
I think that was the case on previous releases before the introduction of UTM Server in the Windows menu, see screenshot for reference:
image

As the issue is gone now, closing the PR.
Thank you @osy

osy:
Ah I see, that's a weird coincidence lol.

PR #5985, opened 2023-12-21
scripting: Add displays to create and config.

connyay:
attempting to script a windows vm with a display


connyay:
I would like to thread through the `hardware` qemu property also

naveenrajm7:
+1 for this feature.

If this is working, I'd appreciate an example usage in apple script format So I can test it.

osy:
Superseded by #6978

PR #6115, opened 2024-02-25
Introduce UTM Remote client for iOS/visionOS

osy:
- [x] Add UTM Remote server functionality for macOS
- [x] Add UTM Remote client functionality for iOS
- [x] Implement UI for picking server and connecting
- [x] Refactor `UTMQemuVirtualMachine` to `UTMSpiceVirtualMachine` and `UTMRemoteSpiceVirtualMachine`
- [x] Re-introduce socket support for SPICE client
- [x] Support QEMU monitor/guest agent through named pipes (because we cannot use remote SPICE pipe to control QEMU)
- [x] UPnP/NAT-PMP for WAN connections
- [x] Synchronize VM states across server and multiple clients
- [x] Support re-connection of SPICE session and UTM Remote session
- [x] Support multiple sessions on iOS and visionOS


wyattanderson:
@osy are there plans to allow the Mac OS app to function as a generic SPICE client? Would be neat to be able to use this to connect to SPICE-capable VMs hosted on Proxmox, for example.

osy:
I think it would never be on the roadmap because it would be too much work. But most of the SPICE functionality is self contained in https://github.com/utmapp/CocoaSpice and it wouldn't be too much work to make a minimal client using it.

PR #6118, opened 2024-02-26
`package_mac.sh`: fix fallback to `hdiutil`

reidpr:
Prior to this PR, if the test for `appimg` on line 118 failed, the script would immediately exit (because of `set -e` on line 3), without running the test on line 119. I.e., the script would just stop, producing neither a `.dmg` nor an error message.

This PR moves the test into a subshell in the `if`, and now the intended `hdiutil` fallback works.


reidpr:
Thanks. I can’t use square brackets because we’re testing the result of a command, rather than a conditional, but I can remove the parentheses entirely.

Squashed and force-pushed.

osy:
Makes sense, thanks

PR #6119, opened 2024-02-26
WIP: enable suspend for QEMU hypervisor on x86-64?

reidpr:
This PR attempts to enable suspend with the QEMU hypervisor on x86-64.

I'm suspicious because it seemed way too easy, but I thought worth sharing anyway. It apparently works for me on my 2019 i7 MacBook Pro running 14.3.1.

Clicking on the close box of a running VM causes the console window to show a throbber for a little while, then close. I can then quit UTM, restart UTM, and click “play” on the VM. The VM console window opens, throbs for a bit, appears to have resumed in the same place, and is responsive.

This VM has no network devices b/c I don’t have the right Apple Developer magic.


reidpr:
Something is goofy; I’m seeing a fair number of “Timed out waiting for RPC”.

I don’t have much time to dig into this, unfortunately.

That said, here is a [debug.log](https://github.com/utmapp/UTM/files/14411794/debug.log) that starts from stopped, suspends, resumes, and then stops. The video device is “VGA” in the GUI. It does say `-accel hvf`.

Issue #4190, opened 2022-07-06
Guest to guest network for multiple guests


dsxack:
**Describe the issue**
I need to build network between 3 guests:

- pfSense firewall (x86, with 2 network adapters: WAN shared network with host and LAN network adapter as gateway for guests, DHCP on LAN network adapter)
- kali linux (arm64, 1 network adapter in LAN)
- metasplotable (x86, 1 network adapter in LAN)

So, I tried to add next QEMU flags to my guests configuration:

- `-netdev socket,mcast=230.0.0.1:1234` not working (VMs starting but unreachable)
- `-netdev socket,connect/listen` working! But just for 2 guests (1 to 1 guest networking)
- `-netdev vde` QEMU in UTM compiled without VDE support
- `-netdev tap` not supported on MacOS as I know

Someone know another ways to build one network with 3 guests?

**Configuration**
* UTM Version: 3.2.4 (58)
* OS Version: 12.5 Beta (21G5046c)
* Intel or Apple Silicon? Apple Silicon

dsxack:
Reached with:

VM1:
`-netdev vmnet-host,id=vm1,net-uuid=8493bb04-56aa-4ec0-a937-605ac1a8bd07`

VM2:
`-netdev vmnet-host,id=vm2,net-uuid=8493bb04-56aa-4ec0-a937-605ac1a8bd07`

VM3:
`-netdev vmnet-host,id=vm3,net-uuid=8493bb04-56aa-4ec0-a937-605ac1a8bd07`

PR #6186, opened 2024-03-10
Initial support for external/multiple networks in vmnet host mode

Schamper:
This adds support for connecting QEMU VMs to specific host networks using the `net-uuid` option (described [here](https://www.qemu.org/docs/master/interop/qemu-qmp-ref.html#qapidoc-1445)). This can be extremely useful when connecting multiple VMs together, or for connecting VMs across virtualisation applications (VMware Fusion/Parallels). Unfortunately, I didn't find such option for the Apple virtualisation APIs, so it's only supported for QEMU VMs. #4190 essentially accomplished this by manually editing the QEMU arguments.

Note that using a custom network UUID will disable the DHCP that macOS vmnet would otherwise provide. As of right now, the user is responsible for providing their own DHCP on the network. E.g. by running a VM that provides it, or running a VM in e.g. VMware on the same network.

I've also added a button for importing a VMware Fusion networking configuration. It requires you to browse to and select `/Library/Preferences/VMware Fusion/networking`. All defined networks in there will be added. As mentioned before, I'm not aware of a more idiomatic approach for this. Adding support for other virtualisation applications is possible, but Fusion is my personal daily driver so that's what I know and have.

This is my first time writing any Swift code, so please excuse the code quality. I only started reading the basics of Swift a few days ago, so I'm not aware of any of the idiomatic or recommended ways of achieving certain things. I'll take any suggestions on that front.

### DIFF
--- a/Configuration/UTMConfigurationHostNetwork.swift
+++ b/Configuration/UTMConfigurationHostNetwork.swift
@@ -1,0 +1,124 @@
+//
+// Copyright © 2024 osy. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


yunzheng:
Any update on getting this merged? I would love to be able to connect to VMware Fusion networks.

Issue #5938, opened 2023-12-01
Update obsolete OS icons/logos?


rxhfcy:
I was looking at the list of OS icons and noticed that some of the projects have since changed their icons after the icons in UTM were last updated.

I then decided to quickly check each icon. According to my very cursory "research", here are my findings (these should of course be verified by someone else, the screenshots mostly use Wikipedia articles for the current icon, or if I found a "more authoritative" source, I included that one instead)

Here are links to the screenshots (each screenshot shows the icon currently in UTM / what I guess is the current icon).

These are the ones I think someone might actually care about (the rest are IMHO pretty obscure and/or obsolete but what do I know):
- [Arch Linux (subtle difference in the logo)](https://github.com/utmapp/UTM/assets/42669366/f5ba706a-6cf3-45fc-8ee3-72b081c1c813)
- [Ubuntu](https://github.com/utmapp/UTM/assets/42669366/fdeaa46e-81ee-4819-ab72-c706accf73d3)
- [Linux Mint](https://github.com/utmapp/UTM/assets/42669366/25c86625-d39f-41a7-99f9-3422c77f9ef5)
- [Lubuntu](https://github.com/utmapp/UTM/assets/42669366/f3e04a0c-44f9-44c3-8d3a-6d02e1551ceb)
- [Android](https://github.com/utmapp/UTM/assets/42669366/959b5185-0cc5-419b-b8ed-0eea4628f54c)
- [Red Hat](https://github.com/utmapp/UTM/assets/42669366/7eeed5e7-2365-4b9e-8443-3a945e057c46)
- [Windows 98 (pretty subtle)](https://github.com/utmapp/UTM/assets/42669366/83cfc1dc-5cbd-45e5-925d-84fbc3b1d6ab)
- [Slackware](https://github.com/utmapp/UTM/assets/42669366/3f5133a3-4102-4dbf-97ce-7530392bc39c)
- [FreeBSD](https://github.com/utmapp/UTM/assets/42669366/d53c997d-da85-4cc1-bbfc-34b5fce71eb9)
- [NetBSD](https://github.com/utmapp/UTM/assets/42669366/2f23a288-e2c1-46d5-be67-8e5df7b1d51a)
- [Apple TV](https://github.com/utmapp/UTM/assets/42669366/771ab2b3-d803-4ca1-adb1-36ef300d18bd)

The rest, for completeness sake:

[...]

PR #6196, opened 2024-03-15
Update obsolete OS icons/logo

mohsenbakhit:
### Adds: 
- New updated icons for some OS's whose logos have changed since.

Addressing #5938


tifasoftware:
The old logo for windows9x is visually better than the proposed.

mohsenbakhit:
@tifasoftware I did think so too. Any other changes you would suggest?

mohsenbakhit:
@rxhfcy I went back and checked, and you are right. Some of the photos did seem squished so I am redoing them to make sure the original aspect ratio is intact while fitting in a 128px square. 

With the android logo should I instead use the 3d logo that was announced more recently?

Also about the windows9x, I believe the titled icon was officially used by Microsoft for the release of 95 and 98, should I instead add that icon?

mohsenbakhit:
@osy I think this PR is merge-ready.

PR #6268, opened 2024-04-12
Re-grab input when automatic capture is enabled

### LINKED ISSUE #6242: UTM doesn't re-grab input when swiping back and forth in fullscreen.
Even though "capture input automatically when entering full screen" is selected in settings, when I swipe from the fullscreened UTM guest to my host Mac desktop (using three fingers on the trackpad), and then back to the guest, the input does not get re-captured.

**Configuration**  
* UTM Version: Version 4.4.5 (94)
* macOS Version: 14.4.1 (23E224)
* Mac Chip (Intel, M1, ...): M3 Max

haroldm:
Make the window auto-capture the mouse again in full screen when it regains focus (for instance after swiping back to the UTM app).

fixes #6242

### DIFF
--- a/Platform/macOS/Display/VMDisplayQemuMetalWindowController.swift
+++ b/Platform/macOS/Display/VMDisplayQemuMetalWindowController.swift
@@ -403,7 +403,6 @@ extension VMDisplayQemuMetalWindowController {
     func windowDidEnterFullScreen(_ notification: Notification) {
         isFullScreen = true
         if isFullScreenAutoCapture {
-            captureMouseToolbarButton.state = .on
             captureMouse()
         }
     }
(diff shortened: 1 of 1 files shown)


Platform/macOS/Display/VMDisplayQemuMetalWindowController.swift
```diff
@@ -403,19 +403,24 @@ extension VMDisplayQemuMetalWindowController {
     func windowDidEnterFullScreen(_ notification: Notification) {
         isFullScreen = true
         if isFullScreenAutoCapture {
-            captureMouseToolbarButton.state = .on
```
osy, line 406: Why did you remove this line and 414?

haroldm: I think these lines are already present in captureMouse() and releaseMouse(). I just tried to check the behavior but I don't think the toolbar is ever visible in full screen when the mouse is captured.

PR #6274, opened 2024-04-14
display(macos): add a setting to start a given VM in fullscreen

### LINKED ISSUE #4304: Option to open vm full screen automatically
Currently Virtual machine windows open on windowed mode, but lots of people like to use it on a fullscreen mode for immersive experience. It would be great to have an option to chose the default behavior.

haroldm:
I didn't add the setting in the display section because it will toggle all windows (displays and terminals) to full screen.
I'm open to changing the parameter location and/or the behaviour.

fixes #4304


osy:
Instead of a setting, we will now retain if the VM was last run in fullscreen or not.

PR #6288, opened 2024-04-19
Stringfix: Macintosh is not MediaAccessControl

elosha:
Fixed "PowerMAC" strings 😄


osy:
As noted in the file header, these files are automatically generated from the output of QEMU. If it is changed in https://gitlab.com/qemu-project/qemu then we will update the file.

elosha:
Ok thanks, will try it over there.

Issue #6428, opened 2024-06-16
[FR] implement iOS 18/OSX Sequoia dark icons


user858753257:
Would be a good addition when UTM (also the iOS remote app) could implement the dark/tinted icons on both platforms to have a streamlined look&feel. 

More Infos :
https://developer.apple.com/design/human-interface-guidelines/app-icons#iOS-iPadOS

![image](https://github.com/utmapp/UTM/assets/8161283/34f53f32-33ba-4449-8e66-25d2021070a0)

wannafedor4:
The tinted icon is already working for me

user858753257:
@wannafedor4 ok i. have adjusted the title.  Tinting is working . But the dark one is missing (black edge colored logo)

wannafedor4:
I made an icon that follows [Apple's documentation on dark app icons](https://developer.apple.com/design/human-interface-guidelines/app-icons)
![UTM Icon](https://github.com/utmapp/UTM/assets/151413500/ab082d38-1bcf-4748-a5fd-ed1a56dacfca)

user858753257:
Nice 👍🏽
When the Owner is also happy we can close here

wannafedor4:
@user858753257 
> Nice 👍🏽
> 
> When the Owner is also happy we can close here

Will open a pull request when I get back to my Mac.

wannafedor4:
#6522 has been created

PR #6522, opened 2024-07-28
Add iOS 18's Dark Mode App Icon Asset

itjustcrashed:
This is from #6428 which requests support for iOS 18's dark icon. Simply added a new image to the `AppIcon` asset catalog that is a transparent, color inverted version of the UTM logo.


wannafedor4:
Are there currently any problems with this change?

osy:
Dark icons have been added in dcdf807851f175886866b76e26097a663284802f

PR #6562, opened 2024-08-16
Added Arabic

muhammadbahaa2001:


muhammadbahaa2001:
Arabic Localization for iOS soon

PR #6617, opened 2024-08-23
Update Japanese localization

MMP0:


osy:
Hi @MMP0, there's a couple of new strings so I will wait for your update before finalizing the next release. Thanks.

MMP0:
@osy Thank you for waiting. It’s okay to merge.

PR #6716, opened 2024-10-04
scripting: add export command

### LINKED ISSUE #6553: Add utmctl export command
## What 

```bash
utmctl export <identifier> <export-path>
```

## How 

There is already a 'Share...' action that exports the VM and its contents, we need to expose this function via utmctl.

## Why 

This allows us to programmatically export the VMs.
Plugins like [Vagrant-UTM](https://naveenrajm7.github.io/vagrant_utm/ ), [Packer-UTM](https://github.com/naveenrajm7/packer-plugin-utm) can use this API to package UTM VMs. 
Useful for packaging and distributing custom UTM VMs.

Related #3718

naveenrajm7:
Fixes #6553 

* Exposed VM export via Apple Scripting and CLI
* Turned on sandbox for unsigned utmctl binary to be consistent with unsigned UTM. So we can test the same data across unsigned products.

Example: 

Usage in AppleScript

```applescript
tell application "/Users/***/UTM.app"
  set vm to virtual machine named "ubuntu"
  set output to POSIX file "/Users/***/ubuntuExport.utm"
  export vm to output
end tell
```

Usage in cli (path must be writable by utmctl)
```bash
utmctl export ubuntu --path /Users/***/ubuntuExport.utm
```

### DIFF
--- a/Scripting/UTM.sdef
+++ b/Scripting/UTM.sdef
@@ -92,6 +92,15 @@
             </parameter>
         </command>
         
+        <command name="export" code="coreexpo" description="Export a virtual machine to a specified location.">
+            <cocoa class="UTMScriptingExportCommand"/>
+            <access-group identifier="*"/>
+            <direct-parameter type="virtual machine" requires-access="r" description="The virtual machine to export."/>
+            <parameter name="to" code="efil" type="file" description="Location to export the VM to.">
+                <cocoa key="file"/>
+            </parameter>
+        </command>
+        
         <class name="virtual machine" code="UTMv" description="A virtual machine registered in UTM." plural="virtual machines">
           <cocoa class="UTMScriptingVirtualMachineImpl"/>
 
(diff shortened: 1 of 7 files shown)


osy:
I'm curious how to document the fact that `utmctl` can only specify paths inside the sandbox. Any ideas?

osy:
Maybe it would be better to not allow passing absolute paths but only relative paths?

osy:
Can you explain what the use case for `utmctl` to only export (and import) to paths that are in the sandbox? I'm wondering if this should be in utmctl at all now... I do recall explicitly not adding this functionality for this reason.

naveenrajm7:
Thanks for your comments, with these and the previous discussion on file URLs now I understand the problem.

When I did this pull request, I did not realize that utmctl command would not work for all paths. 
This is because of the wrong test that I did. As shown in example, I first tried export in applescript and then used the same file path after deleting the previous export file to test in utmctl , it worked. But now I see any other file path tried directly via utmctl will not work.

We have two options,
1) Try to get access to path inside utmctl. Similar to what `POSIX file` in applescript does for us to use that file.
2) Remove support for export command from utmctl **only**. ( import is not in utmctl)

From your comments, I get a sense that option 1 is not feasible or desirable. 
I will remove export from UTMCtl file, please acknowledge if you agree. 

Note: Using these commands via apple script will still work, because the POSIX file of applescript takes care of permissions for us.

naveenrajm7:
Thanks for agreeing to remove it by yourself. 

The Xcode build from main branch fails for me with below error.
```
ld: framework 'phodav-3.0.0' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

osy:
Yes you need to download the latest dependencies from the actions

Issue #6612, opened 2024-08-22
Add utmctl import command


naveenrajm7:
## What 

Import a UTM file into UTM and return UUID. Also, reports progress of the import through percentage/bars.

```bash
> utmctl import -name <VM-NAME> <UTM-FILE> 
Importing ... 0%
Importing ... 50%
Importing ... 100%
virtual machine id <UUID>
```

**NOTE**: 
import is different from 'Open with UTM' , Open with UTM which is already present in Scripting Bridge just creates a short cut in UTM. However in terms of IaC, when we delete a VM user expects all the data of that VM to be also deleted, but for shortcut VMs , delete will only delete the shortcut not the data.

## How 

There is already a command `downloadVM` used as 'downloadVM?url=...' action that imports the UTM zip file as VM into UTM. It also reports a progress bar of the import.  we need to modify it to handle local files and expose this via utmctl.

## Why 

Using `downloadVM` If we want to spin up multiple VMs using the same UTM file, we currently download the file from internet each time for every VM.
This allows us to programmatically import the VMs using local files. 
This will also allow us to use Vagrant Boxes, so we can build, publish, share and use custom UTM VMs using [Vagrant Cloud](https://developer.hashicorp.com/vagrant/vagrant-cloud)

Plugins like [Vagrant-UTM](https://naveenrajm7.github.io/vagrant_utm/ ) can use this API to import VMs and support the Vagrant Box format. 

Related #3718

thegushi:
The applescript is already there to do this.  Could the "Cheat Sheet" (https://docs.getutm.app/scripting/cheat-sheet/) be updated to include a functional example as to how to do this?

PR #6734, opened 2024-10-10
scripting: add import command

naveenrajm7:
Solves (partially) #6612 .

* Added new function importNewUTM  
* Expose importNewUTM as an **import** command in Apple Script.

> if useful the same function can be exposed via UI (File > Import)

Apple Script usage

```applescript
tell application "/Users/x../UTM.app"
  set vmFile to POSIX file "/Users/x../Desktop/ubuntu.utm"
  set vm to import new virtual machine from vmFile
end tell
```

Returns VM UUID 
```bash
%osascript import_vm.applescript
virtual machine id D9C7A2A9-D1C9-4F8D-A733-2E028C573234
```

Import is not yet exposed through utmctl , this will be added in future.
Reason: The bridge protocol generator produced a function signature 
`importNew(_ new_: NSNumber!, from: URL!) -> SBObject` , I don't know how to pass 'virtual machine' ObjectClass as NSNumber. While this is being figured out, just having import function (multiple VMs with same utm file) will enable lot of other things (like vagrant box support, see https://github.com/naveenrajm7/vagrant_utm/discussions/3).


Scripting/UTM.sdef
```diff
@@ -92,6 +92,17 @@
             </parameter>
         </command>
         
+        <command name="import" code="coreimpo" description="Import a new virtual machine from a file.">
+            <cocoa class="UTMScriptingImportCommand"/>
+            <parameter name="new" code="imcl" type="type" description="Specify 'virtual machine' here.">
```
osy, line 97: Feels weird to have a parameter that doesn't do anything and also fails if you pass in another type. Is is possible to rename the command to something like "import virtual machine" and have only one parameter?

naveenrajm7: This is similar to 
`make new virtual machine with properties ...`

Since both import and make creates new object (NSobject), I have used same syntax and code. 

The new keyword is a convention in Apple script to indicate creation of new object and clearly implies you are creating and importing a new virtual machine in UTM. 

Also, this command will always create new VM, rather than VM short cut.

Issue #5211, opened 2023-04-19
Please provide mechanisms for launching unattended OS installations in VMs, especially Linux kickstart/preseeding


mprzybylski:
The feature must provide some automated way of "typing" on the console after the VM boots.

The feature should also provide some way of launching an ephemeral web server and related infrastructure that allows a guest to retrieve an unattended installation settings file.

HashiCorp Packer's qemu builder includes good examples of both features: https://developer.hashicorp.com/packer/plugins/builders/qemu

osy:
Have you tried https://docs.getutm.app/scripting/cheat-sheet/

mprzybylski:
> Have you tried https://docs.getutm.app/scripting/cheat-sheet/

I have, @osy.  While it covers the creation of the vm, it does provide any mechanisms for interacting with the guest OS at boot time, nor does it provide an easy way to supply the guest OS with a kickstart or preseed file.

brad-x:
This can be accomplished with an ISO image or attached USB device containing a kickstart/cloud-init/whatever definition file - is the request here to create a UI for that?

mprzybylski:
> This can be accomplished with an ISO image or attached USB device containing a kickstart/cloud-init/whatever definition file 

It can, but the process of creating that file is really finicky, which is why I asked for functionality that looks like the way Packer's qemu builder serves up kickstart files via an ephemeral http server.

_**But even more importantly**_, UTM's scripting API needs to provide a way to "type" on the VM's console once the VM "powers on."  Some installer images like Ubuntu need `autoinstall`[^1] appended to the kernel command line in order to make proper use of the kickstart file, however it gets supplied.

[^1]: As noted in https://ubuntu.com/server/docs/install/autoinstall-quickstart, "Unless you interrupt boot to add autoinstall to the kernel command line, the installer will prompt for confirmation before touching the disk."

stephan-devop:
I can only second this feature request!

If we could create packer images for UTM a whole new world would open up. I'd like to use packer with ansible to create UTM packages like the derivative-maker does for the Whonix UTM template.

jnoxon:
Has anyone tried this? https://github.com/naveenrajm7/packer-plugin-utm

naveenrajm7:
Hello! 
I am the developer of [UTM packer plugin](https://github.com/naveenrajm7/packer-plugin-utm).
Currently, the plugin supports building custom UTM VM with provisioner support  using an already existing UTM file. (uses builder utm-utm)

[...]

PR #6741, opened 2024-10-12
scripting: add qemu additional arguments

### LINKED ISSUE #6661: Scripting: Add qemu additional arguments
## What 

We can already add qemu arguments in UI, This request is to allow scripting interface to update qemu additional arguments of **qemu configuration**.

## How 

In [UTMScriptingConfigImpl.swift](https://github.com/utmapp/UTM/blob/main/Scripting/UTMScriptingConfigImpl.swift) add functions to update  `additionalArguments: [QEMUArgument] = []` of `UTMQemuConfigurationQEMU`

## Why

Exposing API to update qemu additional arguments allows script/plugins to launch VM with custom arguments.
Examples:
* Launch VM with VNC session to do automated OS install
* Launch VM with specific host path for directory sharing.

naveenrajm7:
Fixes #6661 

read and update qemu additional argument of vm config

Usage:

```applescript
tell application "/Users/xxx/UTM.app"
  set vm to virtual machine named "VM"
  set config to configuration of VM
  
  -- read  qemu additional arguments 
  set qemuAddArgs to qemu additional arguments of config

  -- add
  set newArg to {{argument string:"-vnc 127.0.0.1:1"}}
  set qemuAddArgs to qemuAddArgs & newArg
  
  -- update 
  set qemu additional arguments of config to qemuAddArgs
end tell
```

This will enable to launch VMs with custom qemu args via APIs
Related #5211

### DIFF
--- a/Scripting/UTM.sdef
+++ b/Scripting/UTM.sdef
@@ -425,6 +425,11 @@
             description="List of serial configuration.">
             <type type="qemu serial configuration" list="yes"/>
           </property>
+          
+          <property name="qemu additional arguments" code="QeAd"
+            description="List of qemu arguments.">
+            <type type="qemu argument" list="yes"/>
+          </property>
         </record-type>
         
         <enumeration name="qemu directory share mode" code="QeSm" description="Method for sharing directory in QEMU.">
@@ -532,6 +537,14 @@
               
             <property name="port" code="PoRt" type="integer"
               description="The port number to listen on when the interface is a TCP server."/>
+        </record-type>
+        
+        <record-type name="qemu argument" code="QeAr" description="QEMU argument configuration.">
+          <property name="string" code="StAg" type="text"
+            description="The QEMU argument as a string."/>
+            
+          <property name="file urls" code="FlUr" type="file" list="yes"
+            description="Optional URLs associated with this argument."/>
         </record-type>
         
         <record-type name="apple configuration" code="ApCf" description="Apple virtual machine configuration.">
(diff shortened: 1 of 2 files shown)


Scripting/UTM.sdef
```diff
@@ -534,6 +539,14 @@
               description="The port number to listen on when the interface is a TCP server."/>
         </record-type>
         
+        <record-type name="qemu argument" code="QeAr" description="QEMU argument configuration.">
+          <property name="argument string" code="ArSt" type="text"
+            description="The QEMU argument as a string."/>
+            
+          <property name="file urls" code="FlUr" type="file" list="yes"
```
osy, line 546: Is this actually useful? UTM makes no attempt to open up the sandbox for custom args which contains paths meaning that any path you pass in that's not an absolute path to something that it can already access will fail.

naveenrajm7: Thanks for the explanation.
I kept it to accommodate all the fields of a Qemu Argument type variable. 

Since it is not used currently (permissions need to be taken care if we use it), I will remove it.

naveenrajm7: Dropped support for file urls in qemu arguments in scripting interface.
( Side note: I initially kept it, with a view that we might  use it in future to programmatically configure directory sharing. Can I achieve that by simply passing directory path in args string or it has to be a URL..? Something to test.)

Thanks for reviewing.

osy (approved):

Issue #6914, opened 2024-12-22
scripting: add icon to config


naveenrajm7:
While building VMs, it would be convenient to automatically update VM icons.

Goal 1: Ability to select built-in icons via icon key (string)
Goal 2: Ability to send custom icons via image file (URL)

PR #6916, opened 2024-12-23
scripting: expose icon in config

naveenrajm7:
* Allows to get/set built-in icons 
iconURL is URL in UTMConfig, but stored as string in config.plist, same is done for scripting interface.

Usage:

```applescript
tell application "UTM"
  set vm to virtual machine named "OpenBSD"
  set config to configuration of vm
  set icon of config to "openbsd"
end tell
```

>Say hi to puffy!

For reference :
All available icons are in https://github.com/utmapp/UTM/tree/main/Icons
corresponding icon keys in https://github.com/utmapp/UTM/blob/8ef19ff661c628f3e9075de81f3051471d54de10/Platform/Shared/VMConfigInfoView.swift#L321

### DIFF
--- a/Scripting/UTM.sdef
+++ b/Scripting/UTM.sdef
@@ -405,6 +405,9 @@
         <record-type name="qemu configuration" code="QeCf" description="QEMU virtual machine configuration.">
           <property name="name" code="pnam" type="text"
             description="Virtual machine name."/>
+        
+          <property name="icon" code="IcOn" type="text"
+            description="Virtual machine icon."/>
             
           <property name="notes" code="NoTe" type="text"
             description="User-specified notes."/>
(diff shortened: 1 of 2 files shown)


naveenrajm7:
Fixes #6914 

For future :
If you want to allow custom icons through scripting,
* Take icon as URL from scripting,  
* Expose isIconCustom field in config
* Take care of sandbox permissions of the given icon file
* Copy icon file to VM bundle, set URL in config accordingly
...
> not worth it

NothinRandom:
Hello,

What is the correct syntax to use a custom icon for scripting?  I tried getting the icon file as POSIX file with this isIconCustom (assuming bool), but UTM throws error on invalid config

Thanks

naveenrajm7:
@NothinRandom  as noted in the previous comment, I have NOT added support for using custom icon through scripting interface.
In order use an icon file outside UTM app sandbox you'd need to do complicated sandbox dance in the code to get that file inside the sandbox.

NothinRandom:
@ naveenrajm7 Sorry, totally misinterpreted your message.  Looking forward for this feature when it's available!

PR #6977, opened 2025-01-19
scripting : add vm registry suite

### LINKED ISSUE #6899: scripting: add sandbox permission to custom path through applescript
## What 

Need the ability to pass a custom path with access permission to qemu via additional arguments.

## Why

Currently, any qemu argument with a custom path will fail with error 'Operation not permitted'.
Although, applescript allows you to start a scoped access for a given path in an app's context, doing this in UTM app will not trickle down to qemu, because, qemu has its own app sandbox separate from UTM main app. (UTM qemu does not have its own scripting interface)

*Example usage for custom path in qemu: Multiple Shared Directories*

```
-fsdev local,id=virtfs2,path=/mypath,security_model=mapped-xattr 
-device virtio-9p-pci,fsdev=virtfs2,mount_tag=share2
```

Users who need to share multiple directories via virtfs can do so if they pass multiple virtfs arguments by specifying path.
This works only if qemu can access the specified path. 

Alternative:
https://github.com/utmapp/UTM/discussions/5463#discussioncomment-11478233 - Currently users are resorting to making UTM create a sandbox permission for a given path through Shared Directory 'Browse..' option (The only way to open a path to sandbox) and then using it in the qemu argument. This could be made simpler.

Side Note: 
Opening up qemu additional args via scripting have been very useful. eg: 
1. launching vnc session via `-vnc`
2. Using UTM as a private cloud , by passing custom cloud-init files via `-smbios` to fire up UTM cloud images.
3. Share multiple directories ...

## How

There are already functions in place to open a path to qemu via the "complicated sandbox dance" (Bookmark in UTM -> SecureScope in QEMUHelper (Also available to QEMULauncher) -> Store back the Bookmark in UTM for future use by QEMU ). we need to expose this to user in some form so users can pass their own path and use it with qemu.

There are two ways to expose :
1. Use existing files parameter in qemu additional arguments
```applescript
set qemuAdditionalArgs to {{argument string:"-fsdev local,id=virtfs1,path=/mypath", file urls:{"/mypath"} }
```

In file urls user should pass the path used in the argument string. While updating the qemu config, UTM should open access to this path and store it. This allows the proper use of file with qemu.

2. Make a separate command to request sandbox permission
if the file urls in qemu args object is meant for something else, then we need a custom command in applescript for users to request permission.

naveenrajm7:
Fixes #6899

The registry suite is used to read and update vm registry.
All configs saved to disk is exposed by Configuration suite, similarly those that are not saved to disk are stored in RegistryEntry of a VM, they will now be exposed by Registry suite.

Users will now be able to give UTM→QEMU access to their custom path by updating registry, which can then be used in qemu additional args.

Added

- UTMScriptingRegistryEntryImpl will control what will be exposed from RegistryEntry to scripting bridge, currently only shared directory is exposed in the form of [URL].
    
    Since we have a separate suite for handling registry , this suite can grow on what will be exposed. Eg: externalDrives  (Not everything in RegistryEntry is suitable to be exposed)
    
- Added file urls to qemu addtional args, QEMU launcher looks at file urls of args to collect required resources. These files urls are not stored to the disk, meaning they will be lost .
    
    However, they exists only for QEMU to figure out permission for these resources. Since the permission should actually come from registry, it is okay to lose these entries in qemu args.

### Usage: Registry Suite

registry_add.applescript
```applescript
on run argv
  set vmName to item 1 of argv
  -- Define the directory path
  set dir1Path to POSIX file "/Users/utm/Dir1"
  set dir2Path to POSIX file "/Users/utm/Dir2"

  tell application "UTM"
      set vm to virtual machine named vmName
      
      -- get current registry
      set reg to registry of vm
      -- add new entries
      set newEntry to {dir1Path,dir2Path}
      set reg to reg & newEntry
      -- update registry  ( No need for the VM to be stopped )
      update registry of vm with reg -- QEMU now has access to Dir1, Dir2
  end tell
end run
```

`osascript registry_add.applescript testVM`

### Example use-case for registry: directory share

```applescript
on run argv
    set vmName to item 1 of argv
     -- share id to support multiple shares and avoid conflict with UTM's default share (id=0)
    set dirId to "1" 
    set dirPath to "/Users/utm/Dir1"
    set dirURL to POSIX file dirPath
    
    -- Prepare the QEMU argument strings
    set fsdevArgStr to "-fsdev local,id=virtfs" & dirId & ",path=" & dirPath & ",security_model=mapped-xattr"
    set deviceArgStr to "-device virtio-9p-pci,fsdev=virtfs" & dirId & ",mount_tag=share" & dirId

    tell application "UTM"
        -- Get the VM and its configuration
        set vm to virtual machine named vmName
        set config to configuration of vm

        -- Get the current QEMU additional arguments
        set qemuAddArgs to qemu additional arguments of config

        -- Add the new arguments to the existing ones
        set end of qemuAddArgs to {argument string:fsdevArgStr, file urls:{dirURL}}
        set end of qemuAddArgs to {argument string:deviceArgStr}

        -- Update the configuration with the new arguments list
        set qemu additional arguments of config to qemuAddArgs
				
	-- Give UTM access to the shared directory by updating registry
        -- Get the current directory shares in registry
        set reg to registry of vm
        -- Add new directory shares to the registry
        set reg to reg & {dirURL}
        -- Update registry of vm with new directory shares
        update registry of vm with reg
	
    end tell
end run
```

Mount inside VM

```
sudo mkdir -p /mnt/share1
sudo mount -t 9p -o trans=virtio,version=9p2000.L share1 /mnt/share1
```

### DIFF
--- a/Scripting/UTM.sdef
+++ b/Scripting/UTM.sdef
@@ -712,4 +712,27 @@
           <direct-parameter description="The USB device to disconnect." type="usb device"/>
         </command>
     </suite>
+    
+    <suite name="UTM Registry Suite" code="UTMr" description="UTM virtual machine registry suite. Use this to update virtual machine registry.">
+           <access-group identifier="com.utmapp.UTM.vm-access" />
+           
+           <class-extension extends="virtual machine" description="Virtual machine registry.">
+               <property name="registry" code="ReGs" access="r"
+                   description="The registry of the virtual machine.">
+                   <type type="file" list="yes"/>
+               </property>
+               
+               <responds-to command="update registry">
+                   <cocoa method="updateRegistry:"/>
+               </responds-to>
+           </class-extension>
+           
+           <command name="update registry" code="UTMrUpDt" description="Update the registry of the virtual machine.">
+               <direct-parameter description="Virtual machine to update." type="virtual machine"/>
+               <parameter name="with" code="UpRg" description="The registry to update the virtual machine. Currently you can only change the shared directory with this!">
+                   <cocoa key="newRegistry"/>
+                   <type type="file" list="yes"/>
+               </parameter>
+           </command>
+    </suite>
 </dictionary>
--- a/Scripting/UTM.sdef
+++ b/Scripting/UTM.sdef
@@ -561,6 +561,10 @@
         <record-type name="qemu argument" code="QeAr" description="QEMU argument configuration.">
           <property name="argument string" code="ArSt" type="text"
             description="The QEMU argument as a string."/>
+          <property name="file urls" code="FlUr" 
+            description="Optional URLs associated with this argument.">
+            <type type="file" list="yes"/>
+          </property>
         </record-type>
         
         <record-type name="apple configuration" code="ApCf" description="Apple virtual machine configuration.">
--- a/Scripting/UTMScriptingConfigImpl.swift
+++ b/Scripting/UTMScriptingConfigImpl.swift
@@ -193,6 +193,10 @@ extension UTMScriptingConfigImpl {
         var serializedArgument: [AnyHashable: Any] = [
             "argumentString": argument.string
         ]
+        // Only add fileUrls if it is not nil and contains URLs
+        if let fileUrls = argument.fileUrls, !fileUrls.isEmpty {
+            serializedArgument["fileUrls"] = fileUrls.map({ $0 as AnyHashable })
+        }
         
         return serializedArgument
     }
@@ -517,7 +521,10 @@ extension UTMScriptingConfigImpl {
         let additionalArguments = records.compactMap { record -> QEMUArgument? in
             guard let argumentString = record["argumentString"] as? String else { return nil }
             var argument = QEMUArgument(argumentString)
-            
+            // fileUrls are used as required resources by QEMU.
+            if let fileUrls = record["fileUrls"] as? [URL] {
+                argument.fileUrls = fileUrls
+            }
             return argument
         }
         // Update entire additional arguments with new one.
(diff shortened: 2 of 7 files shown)


naveenrajm7:
This enables sync folder implementation in plugins (https://github.com/naveenrajm7/vagrant_utm/pull/17)

Usage
```ruby
Vagrant.configure("2") do |config|
  config.vm.synced_folder ".", "/vagrant" # Dir1
  config.vm.synced_folder "src/", "/srv/website" # Dir2
  config.vm.box = "utm/bookworm"
end
```

NothinRandom:
@naveenrajm7,  what if OS doesn't support type 9p?  The following out is supported filesystems

```
# cat /proc/filesystems
nodev sysfs
nodev tmpfs
nodev bdev
nodev proc
nodev cgroup
nodev cgroup2
nodev cpuset
nodev devtmpfs
nodev debugfs
nodev tracefs
nodev securityfs
nodev sockfs
nodev bpf
nodev pipefs
nodev ramfs
nodev hugetlbfs
nodev devpts
nodev autofs
      xfs
      zonefs
nodev nqueue
nodev selinuxfs
nodev pstore
nodev efivarfs
      ext3
      ext2
      ext4
nodev configfs
      fuseblk
nodev fuse
nodev fusectl
nodev rpc_pipefs
      vfat
```

naveenrajm7:
You can explore other options for directory sharing.
https://docs.getutm.app/settings-qemu/sharing/#virtfs

Check this on to make sure your Linux has 9p support.
https://docs.getutm.app/guest-support/linux/#drivers

PR #7018, opened 2025-02-15
Update French Localizable.strings

MichelBaie:
Just a translation fix i've seen when we start virtual machine!

### DIFF
--- a/Platform/fr.lproj/Localizable.strings
+++ b/Platform/fr.lproj/Localizable.strings
@@ -125,7 +125,7 @@
 // UTMVirtualMachine.m
 "Suspended" = "Suspendue";
 "Stopped" = "Arrêtée";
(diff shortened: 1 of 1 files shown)


Quetzacoalt91 (approved):

Platform/fr.lproj/Localizable.strings
```diff
@@ -125,7 +125,7 @@
 // UTMVirtualMachine.m
 "Suspended" = "Suspendue";
 "Stopped" = "Arrêtée";
-"Starting" = "Déamarre";
+"Starting" = "Démarre";
```
Quetzacoalt91, line 128: I don't know if we would have enough room for this variant, but it seems more appropriate: `Démarrage en cours`

Issue #3378, opened 2021-12-09
Add check for updates feature


conath:
It's hard to know if UTM is up to date if it's been downloaded from outside the Mac App Store. (one could watch the GitHub releases, but that requires a GitHub account)

There should be an automatic and manual way of checking for updates. 

In the Mac App Store version, the feature could either be excluded or we could have it link to `macappstore://showUpdatesPage?scan=true`, which opens the Mac App Store and checks for updates.

apfimagination:
If you install it with "Brew", all you need to do is run "Brew update UTM"

Francesco146:
any news on this?

PR #7036, opened 2025-02-25
feat(DRAFT): implement update checker and alert for new versions

Francesco146:
This PR introduces an update alert mechanism for UTM on macOS. When a new release is available on GitHub, users will be notified with an alert, allowing them to download the latest version.  

solves: #3378 

this is my first time dealing with swift, so... yeah. i need to test this. i got the app running in xcode locally, but i'm having an hard time following the flow of execution. still, i wanted to put a starting base since the feature is quite usefull and the issue is old-ish.

### DIFF
--- a/Platform/macOS/UTMApp.swift
+++ b/Platform/macOS/UTMApp.swift
@@ -32,6 +32,7 @@ struct UTMApp: App {
                     data.showErrorAlert(message: message)
                 }
             }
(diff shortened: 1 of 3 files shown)


Francesco146:
jfyi i'm still working on this, i figured out how to actually build and run my code... macos development is another level

Issue #7026, opened 2025-02-18
Allow control of shared directory name in MacOS guests.


tmc:
Some programs don't handle spaces in paths appropriately even despite symlinks being in place.

It'd be helpful if the name of the directory were a possible option for macOS guests.

PR #7051, opened 2025-03-11
macOS: Add shared directory mount tag customization

tmc:
This change allows users to customize the mount tag name for shared directories in macOS guests. Some applications don't handle spaces in paths properly, so this provides more flexibility.

- Added sharedDirectoryMountTag property to UTMAppleConfiguration
- Updated shareDirectoryTag to use custom mount tag if set
- Added UI in VMConfigAppleSharingView to modify the mount tag

Addresses #7026 

Screenshot 2025-03-11 at 3 02 06 AM

### DIFF
--- a/Configuration/UTMAppleConfiguration.swift
+++ b/Configuration/UTMAppleConfiguration.swift
@@ -36,6 +36,9 @@ final class UTMAppleConfiguration: UTMConfiguration {
     @Published private var _networks: [UTMAppleConfigurationNetwork] = [.init()]
     
     @Published private var _serials: [UTMAppleConfigurationSerial] = []
+    
+    /// Custom mount tag for shared directories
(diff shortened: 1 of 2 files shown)


osy:
This feels like settings creep (where we eventually will have hundreds of settings and a new user has no idea what is important). Do you have a justification of why this is something that needs to be done in UTM (rather then, say, not have a space in the directory you are mounting)?

tmc:
I feel you, perhaps there could be some advanced settings or cli based way to change the tag -> a non-trivial amount of software does not handle spaces well -- that's the only rationale.

osy:
If it's just the default tag being the issue, I could see a global toggle or checkbox that will select an alternative tag. Does this have to be a setting that is saved per VM?

tmc:
It does impact per vm behavior but I personally would be satisfied with a
global setting.

PR #7253, opened 2025-07-02
qemu patches for virtio-ramfb from v4.6.4

gbaranski:
Hey!
I am trying to run UTM v4.6.5 with the sysroot from v4.6.4.
Unfortunately though all the artifacts are now expired :/ See: https://github.com/utmapp/UTM/actions/runs/13506502834.

I was trying to build it on my own machine, failed though, and I don't really have enough computational power on my MacBook Air to be able to build it myself in a VM.
Well then I was trying to do it with Github Workflows, however it looks like you're using a self hosted runner, which I don't really know how did you configure.

That's why, I was wondering, could I possibly run this PR workflows in your environment, just so that I'd be able to get the sysroot for ARM64 MacOS?

That'd be really amazing, I have been struggling with building it myself for hours already.

Maybe a good idea would be to put these artifacts in the release? So that they don't expire.

Thanks again.


gbaranski:
@osy could you take a look, please :)

osy:
Sorry I've fixed the CI but it's only building the latest version (beta). It should still work for you though. Closing this as this PR is not relevant.

PR #7287, opened 2025-07-10
Show localized bridge network name when available

Schamper:
This adds the localized interface name in bridge network configuration, which makes them a little easier to recognize.

QEMU configuration:
image

Apple configuration:
image


gabefair:
This would be very useful. Thanks

Issue #3294, opened 2021-11-08
Configure shared networking IP address


GregOriol:
**Describe the issue**
When using `shared` networking mode with macOS (12) and ubuntu (20.04), the IPv4 address that the VM gets on my mac is 192.168.64.2.
However, for some use-cases, I need the address to be something else and something pre-defined (for example when using the VM are a development server and some other service has the IP in its configuration).

How could it be possible to define or change the assigned IP to something else? I've tried many options in the advanced settings of the Network tab, but most are not accepted with the shared mode, and none seem to have an effect on the IP address the guest gets.

I've also tried manipulating vmnet from macOS with `ifconfig` (ipconfig set bridge bridgeX...), but I could only change the IP of the bridge and not the VM.

nb: I cannot use the bridged mode with another DHCP server because I don't have control over it.

**Configuration**
* UTM Version: 2.4.0
* OS Version: macOS 12.0.1
* Apple Silicon

ktprograms:
The shared networking is based on the [vmnet](https://developer.apple.com/documentation/vmnet/) framework. You could take a look at [this](https://github.com/hamishcoleman/macos-vmnet/blob/master/simple.c) example and see if you can adapt it to QEMU. The thing of interest would probably be vmnet_start_address_key, vmnet_end_address_key and vmnet_subnet_mask_key. The online (developer.apple.com) documentation doesn't cover these keys, but if you look in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/vmnet.framework/Versions/A/Headers/vmnet.h`, you can find the documentation comments for everything in vmnet.framework. (You need Xcode installed).

GregOriol:
@ktprograms Is the management of vmnet part of UTM or of qemu or of something else? I can't figure out where this is managed: I have seen that parameters are passed to qemu but not where "vmnet-macos" type is handled from there.

ktprograms:
Well, `vmnet-macos` was implemented into QEMU with https://lists.gnu.org/archive/html/qemu-devel/2021-02/msg04637.html. The vmnet-macos code is in https://github.com/utmapp/qemu/blob/utm-edition/net/vmnet-macos.c.

[...]

PR #7380, opened 2025-08-27
Add support for host networking with Apple Virtualization

Schamper:
This is untested (and thus a draft) because I do not have access to a host machine running Tahoe at the moment (and I couldn't get nested virtualization to work).

In the vmnet documentation I noticed a new feature in Tahoe: https://developer.apple.com/documentation/virtualization/vzvmnetnetworkdeviceattachment
There's not a ton of documentation available, but I assume this works similarly to how you can attach to a specific vmnet with QEMU.

If that is indeed how that API works, then this should allow VMs running with Apple Virtualization to be networked with custom host defined networks (as an extension to https://github.com/utmapp/UTM/pull/6186).

### DIFF
--- a/Configuration/UTMAppleConfigurationNetwork.swift
+++ b/Configuration/UTMAppleConfigurationNetwork.swift
@@ -23,11 +23,13 @@ struct UTMAppleConfigurationNetwork: Codable, Identifiable {
     enum NetworkMode: String, CaseIterable, QEMUConstant {
         case shared = "Shared"
         case bridged = "Bridged"
+        case host = "Host"
         
         var prettyValue: String {
             switch self {
             case .shared: return NSLocalizedString("Shared Network", comment: "UTMAppleConfigurationNetwork")
             case .bridged: return NSLocalizedString("Bridged (Advanced)", comment: "UTMAppleConfigurationNetwork")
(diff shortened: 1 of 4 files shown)


andresmmujica:
How could I test this?  I'm having an issue with UTM related to issue #3294 and I think this may work.

Schamper:
I didn't write this up but I looked into this a little while ago, but unfortunately the APIs are not compatible. Maybe there's some magic possible by fiddling with some serialized objects. I think it should be possible with a network helper process and route the network through that, but that's a little out of my depth at the moment and probably has some performance considerations as well. Or cross our fingers that Apple adds support for this in macOS 27.

yangsfang:
I think you can still connect shared networking on macOS <=15. It just requires you to use [VZFileHandleNetworkDeviceAttachment](https://developer.apple.com/documentation/virtualization/vzfilehandlenetworkdeviceattachment). This project https://github.com/nirs/vmnet-helper implemented a helper to forward packets from the socket to a vmnet network. There is a nice diagram on this page explaining how it works https://github.com/nirs/vmnet-helper/blob/main/docs/architecture.md#native-vmnet-on-macos-26 

But on macOS 26+ you no longer need the helper because of vzvmnetnetworkdeviceattachment.

osy:
Superseded by #7865

PR #7426, opened 2025-09-19
Update README.md

wyatt-wong:
Added warning message that UTM doesn't support macOS in Intel platform.


osy:
We don't overload the readme with a list of everything thing the app doesn't do.

PR #7427, opened 2025-09-20
build(macos): fix QEMU build by disabling SDL and adding build include paths

peterneutron:


gabefair:
Could you provide more details about the error you were getting before your fix?

peterneutron:
OS: macOS 26.0.1 (25A362)
ARCH: arm64 (M3 Pro)


```
./scripts/build_dependencies.sh -p macos -a arm64`

...

ld: warning: building for macOS-11.0, but linking with dylib '/opt/homebrew/opt/sdl2/lib/libSDL2-2.0.0.dylib' which was built for newer version 15.0
Undefined symbols for architecture arm64:
  "_qemu_main", referenced from:
      _sdl2_display_init in ui_sdl2.c.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[4923/6254] Compiling C object libqemu-xtensaeb-softmmu.dylib.p/target_xtensa_core-dsp3400.c.o
ninja: build stopped: subcommand failed.
make[1]: *** [run-ninja] Error 1
make: *** [build] Error 2
```

Could not reproduce the error condition that previously needed the addition of

```
--extra-cflags=-I$QEMU_DIR/build \
--extra-cflags=-I$QEMU_DIR/build/include"
```

with the current master branch.

osy:
See: https://github.com/utmapp/UTM/blob/main/Documentation/MacDevelopment.md#building-dependencies-advanced

We recommend using a fresh VM to build dependencies because QEMU will try to pull in stuff from local homebrew install (even though we try to avoid it with a custom pkg-config). While the SDL issue has a visible error, I've dealt with other issues where it wasn't obvious that the cause is something pulled in from the system homebrew. As such, I don't want to include any "patches" that can fix one specific issue from a tainted build without addressing it in general.

PR #7445, opened 2025-10-04
Update availability check for MenuLabel to support macOS 11.0 or later

iCMDdev:
This PR updates an availability check for a SwiftUI `.titleAndIcon` label, which allows decreasing the minimum deployment to macOS 11.0. 

This allows UTM 4.x to build & run on earlier versions of macOS Big Sur (11.x < 11.3).


iCMDdev:
Note that I haven't changed the minimum deployment-related Xcode files in this PR, yet.

iCMDdev:
Now it should be done.

osy:
Have you tried building and testing it?

iCMDdev:
I followed the instructions [here](https://github.com/utmapp/UTM/blob/4bacaa9d3eea5ad7d16009504a0820511f5d7508/Documentation/MacDevelopment.md) (downloaded `sysroot-macOS-arm64` from GitHub Actions) and succesfully built from Xcode 26 (on macOS 26). I didn't get any build errors, but I haven't tested the resulted app on macOS 11.x < 11.3.

Earlier (before making this change), I had access to a Mac on 11.2.3, and I manually changed the minimum version of the official (latest) binary using `vtool`, which resulted in an `Undefined symbol` crash (in SwiftUI if I recall correctly). I think that was likely caused by the [`.titleAndIcon`](https://developer.apple.com/documentation/swiftui/labelstyle/titleandicon) label style, which is unavailable on macOS versions < 11.3.

Xcode indeed throws this error if I change the minimum deployment on the currrent `main` branch (unmodified) code:

Xcode error

iCMDdev:
Visually, I haven't seen anything else that requires macOS 11.3 or later, other than this, but I could be wrong - though building seems fine.

osy:
If I remember correctly, there was a number of SwiftUI bugs that was fixed in 11.3. I don't recall the full list off the bat, but I would try testing VM creation, config (all the tabs, changing all the options), app preferences, and moving around VMs on the home screen.

iCMDdev:
I'll test and see. Either way, if you consider it's better to keep `11.3` as a minimum target, I think it's fine (and people can / should upgrade either way).

iCMDdev:
Sorry for the delay; unfortunately, my Mac doesn’t support a version that low, so I haven’t been able to test this yet. If anyone is willing to test it, I’d appreciate it. However, as I mentioned earlier, I don’t believe this is an urgent matter, as people should upgrade to later versions regardless.

PR #7484, opened 2025-11-13
Added SUSE logo

dmathern86:
Added the SUSE logo for virtual machine which are running SUSE Linux Enterprise Server or SUSE Linux Micro.


osy:
Hi please use the same standard size as all the other icons.

dmathern86:
> Hi please use the same standard size as all the other icons.

Hi, 
changed logo size to 128x128 px.

Issue #7488, opened 2025-11-16
Linux guest disconnects virtiofs frequently


evan314159:
**Describe the issue**  
Linux guest (Debian Trixie, Fedora) virtiofs disconnects and reconnects for no obvious reason.  Files open on the virtiofs at the time have errors, and if the current working directory is on the virtiofs its handle is also invalid:

ls: cannot open directory '.': No such file or directory

When I notice the VM in this state I can restart apps or change directory to the present working directory, and then it will be fine until the next time.

**Configuration**  
* UTM Version: 4.7.4, Apple Virtualisation
* macOS Version: 26.1
* Mac Chip (Intel, M1, ...): M4 Pro

**Crash log**  
None

**Debug log**  
I do not see anything logged in the guest OS -- no dmesg or journalctl
I don not see anything logged on the host -- no Console.app output for "virtiofs" or UTM errors

**Upload VM**  
It is reproducible across different Linux distributions, macOS versions, hardware
It is reproducible but not on demand.  I don't know the conditions that cause this to happen, but it happens anywhere between multiple times daily to after a few days.  For example it has happened twice this morning.

kraju:
I have exactly the same on M2 Pro with Ubuntu 24.04 and Debian 13 running with Apple Virtualization on macOS Sonoma, Sequoia and currently Tahoe. Tried multiple kernels, but it turned out it was UTM-related for me - last working version is UTM 4.5.4 - since then virtiofs shares are reconnected after some time (I found this due to my bind mounts from virtiofs mounts were invalidated after some time). There are no messages, no errors, just virtiofs underneath seems to be updated.

ghost:
@kraju thank you for the hint that the issue started in v4.5.4, which introduced dynamic resolution.  From that I found a repeatable way to cause the issue: resize the window, which post-4.5.4 causes a registry update that reconnected the virtiofs.

PR: #7494.  Thank you!

PR #7494, opened 2025-11-21
Apple Virtualization: only update shared directories if they changed to avoid unnecessary virtiofs reconnections

ghost:
For Apple Virtualization, only update shared directories if they changed to avoid unnecessary virtiofs reconnections.

Although this issue existed prior to v4.5.4, following v4.5.4, UTM introduced dynamic resolution.  Resizing the window with dynamic resolution is a repeatable way to cause a registry update.  updateConfigFromRegistry() updates the shared directories on every call, causing the virtiofs to reconnect.  As the virtiofs disconnects and reconnects, file and directory handles on the share are invalidated causing guest application errors.

To reproduce bug:
- Apple Virtualization, Linux guest, Dynamic resolution enabled
- Open terminal
- $ cd {shared_directory}
- Resize the window
- $ ls
- Result: ls: cannot open directory '.': No such file or directory

Thank you to @kraju for the hint that the issue started in v4.5.4.

### DIFF
--- a/Services/UTMAppleVirtualMachine.swift
+++ b/Services/UTMAppleVirtualMachine.swift
@@ -968,7 +968,18 @@ extension UTMAppleVirtualMachine {
     }
     
     @MainActor func updateConfigFromRegistry() {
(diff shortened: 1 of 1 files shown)


ghost:
Fixes #7488

PR #7534, opened 2025-12-26
Add Rocky Linux logo

elguero:
This adds a logo for Rocky Linux.


osy:
Hi please use the same standard size icon as all the others. Thanks!

elguero:
Sure thing. What is the standard size?  I looked at several icons that are in the repo. I see the size H x W is 128x128, which is what was uploaded.  I see varying bit depths and h/w resolutions, though.

Thank you for the guidance.

osy:
You're right, my browser inspector lied to me.

PR #7572, opened 2026-01-10
apple: use public availability checks for dynamic resolution

jackbolen:
## Summary
- Use macOS 14+ availability checks for dynamic resolution support
- Simplify resolution restore from polling loop to single delayed call

## Details
This updates the dynamic resolution implementation to use the public [`automaticallyReconfiguresDisplay`](https://developer.apple.com/documentation/virtualization/vzvirtualmachineview/automaticallyreconfiguresdisplay) API on [`VZVirtualMachineView`](https://developer.apple.com/documentation/virtualization/vzvirtualmachineview) (macOS 14+).

Changes:
- Remove `supportsReconfiguration` computed property
- Replace checks with `#available(macOS 14, *)`
- Simplify resolution restore timing

### DIFF
--- a/Platform/macOS/Display/VMDisplayAppleDisplayWindowController.swift
+++ b/Platform/macOS/Display/VMDisplayAppleDisplayWindowController.swift
@@ -27,22 +27,10 @@ class VMDisplayAppleDisplayWindowController: VMDisplayAppleWindowController {
         appleView
     }
 
-    var supportsReconfiguration: Bool {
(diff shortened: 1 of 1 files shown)


osy:
Can you explain what you are trying to achieve here? The reason the logic is so complex is because we have different combinations of: macOS guest (below 14 and at least 14), Linux guests, and macOS hosts (below 14 and at least 14). Only specific combinations work hence the complicated logic.

Issue #4636, opened 2022-11-09
VM scrolling locks up


jtran19:
**Describe the issue**  
When scrolling quickly (flick on a Trackpad) or near the ends (top/bottom) of a page, the UI locks up for 5+ seconds.  Most noticeable in vscode-insiders, but also happens in Google Chrome.

**Configuration**  
* UTM Version: 4.1.0[^1]
* macOS Version: 13 (Ventura)
* Mac Chip (Intel, M1, ...): M1

[^1]: Issue also occurs in 4.0.7, 4.0.8 and 4.0.9, but not as noticeably.  When it does happen in these versions, the lock up is not as easily triggered and does not last as long.

**Upload VM**  
[config.plist.zip](https://github.com/utmapp/UTM/files/9971509/config.plist.zip)

osy:
This is likely an Apple issue and would be difficult to debug on our end.

jdork0:
I think this is being caused by this commit: https://github.com/utmapp/UTM/commit/6d6661247ff09e6638765ac3c223f2fef84f836d

I have macOS 13.1 with a macOS 13.1 guest running on an M2 MacBook Air.  I don't see any problems before release 4.0.6 and if I rebuild the latest code without this change, I no longer see the issue.

jtran19:
FWIW, I am using a bluetooth connected Magic Trackpad.  However, it does also happen to me when I'm using the built-in trackpad on my 2021 MBP 16".

jdork0:
Tried again and I was wrong.  It does still happen with the built in trackpad.  I reverted to UTM v3.2.4 as it has no problems for me.

@osy, any chance the trackpad config can be made optional?  macOS VMs on arm Macs are unusable for me with this issue.

jtran19:
I'm not sure what the added trackpad feature was in v4.  I don't have the Trackpad settings option in my System Settings.

jtran19:
Installed the latest beta (4.2.0) and the scrolling issues have been resolved for me.  Doesn't feel like scrolling in the host OS, but it's not locking up anymore.  However, I'm having the right-click issues described by #4645.  Going on memory...I think UTM v3 had the click issues, but they were resolved for me in v4.0.x and v4.1.x.

Sometimes when I right-click, the following sequence occurs:
1. nothing happens, I right-click again
2. nothing happens, I right-click again
3. context menu appears
4. context menu disappears when I move my cursor onto it to select from the menu
5. I right-click again
6. context menu appears
[...]

PR #7573, opened 2026-01-10
apple: configure both trackpad and USB mouse for macOS guests

jackbolen:
## Summary
Include both [`VZMacTrackpadConfiguration`](https://developer.apple.com/documentation/virtualization/vzmactrackpadconfiguration) and [`VZUSBScreenCoordinatePointingDeviceConfiguration`](https://developer.apple.com/documentation/virtualization/vzusbscreencoordinatepointingdeviceconfiguration) when trackpad mode is selected for macOS guests.

## Details
Apple's documentation for `VZMacTrackpadConfiguration` states:

> To support both macOS 13.0 and earlier guests, set pointingDevices to an array that contains **both** a VZMacTrackpadConfiguration **and** a VZUSBScreenCoordinatePointingDeviceConfiguration object.

With both devices present, the framework can route input through the optimal device, which fixes scrolling latency.

Fixes [#4636](https://github.com/utmapp/UTM/issues/4636)

### DIFF
--- a/Configuration/UTMAppleConfigurationVirtualization.swift
+++ b/Configuration/UTMAppleConfigurationVirtualization.swift
@@ -164,8 +164,12 @@ extension UTMAppleConfigurationVirtualization {
                 vzconfig.pointingDevices = [VZUSBScreenCoordinatePointingDeviceConfiguration()]
                 #if arch(arm64)
                 if #available(macOS 13, *), isMacOSGuest && pointer == .trackpad {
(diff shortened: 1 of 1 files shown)


osy:
Does it really fix #4636? This was what we had originally but it caused scrolling issues on older macOS guests (check the commit history). If we add it back I think there needs to be logic to determine what the guest version is before enabling it.

PR #7613, opened 2026-02-13
fix: remove MTL_DEBUG_LAYER to fix system frameworks in guest VMs on macOS 26.x

### LINKED ISSUE #7229: Bridged network interface doesn't work on macOS 26
I've checked the Issues, haven't found anything similar.

**Describe the issue**  
When I opened a VM that worked on a previous macOS version (15.5), it didn't boot successfully and showed error messages like this:
"QEMU error: QEMU exited from an error: gemu-aarch64-softmmu:
-netdev vmnet-shared,id=net1:
cannot create vmnet interface:
general failure (possibly not
enough privileges)" and then it said "Internal error trying to connect to SPICE server."
When I removed the bridged network interface and restarted the VM it worked fine, and there remained an interface connected through "Shared Network" and the internet connection inside the VM was available 

**Configuration**  
* UTM Version: 4.6.5
* macOS Version: macOS 26
* Mac Chip (Intel, M1, ...):  M1 Pro

**Crash log**  
The UTM app itself didn't seem to crash.

**Debug log**  
[debug.log](https://github.com/user-attachments/files/20738158/debug.log)

**Upload VM**  
The problem seems to happen whatever the guest OS is and whatever architecture it runs on, so I don't think it's related to a specific VM.

samfili:
Fixes #7229

## Problem
`MTL_DEBUG_LAYER=1` and `MVK_DEBUG=1` environment variables in QEMUHelper XPC service on the host break system frameworks (Safari, altool, XCreds) **inside guest VMs** on macOS 26.x.

## Root Cause
Metal Debug Layer on the host interferes with GPU virtualization, causing CFNetwork, WebKit, and Security frameworks to fail in the guest OS.

## Solution
Remove these debug variables while keeping safe alternatives like `MVK_CONFIG_LOG_LEVEL`.

## Testing
Tested on macOS 26.3 host with macOS 26 guest - Safari, altool, and XCreds should work after this fix.

### DIFF
--- a/Services/UTMQemuSystem.m
+++ b/Services/UTMQemuSystem.m
@@ -120,16 +120,15 @@ static int startQemu(UTMProcess *process, int argc, const char *argv[], const ch
         self.mutableEnvironment[@"VIRGL_LOG_LEVEL"] = @"debug";
         self.mutableEnvironment[@"MESA_DEBUG"] = @"1";
         self.mutableEnvironment[@"MVK_CONFIG_LOG_LEVEL"] = @"4";
-        self.mutableEnvironment[@"MVK_DEBUG"] = @"1";
-        self.mutableEnvironment[@"MTL_DEBUG_LAYER"] = @"1";
+        // NOTE: MTL_DEBUG_LAYER and MVK_DEBUG are removed because they can break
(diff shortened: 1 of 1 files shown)


osy:
I have no idea what you're talking about because macOS guests never use the QEMU backend.

Issue #7604, opened 2026-02-01
Better presentation for read-only args


osy:
Too much confusion from users thinking that QEMU arguments should be editable. It's presented for debugging issues but has never been used that way. The debug.log shows the arguments used and we can keep the export arguments for debugging.

sdrabb:
@osy I started with this, let me know what do you think about this.

PR #7621, opened 2026-02-21
[macOS] Remove QEMU arguments table from settings (#7604)

sdrabb:
Remove the read-only QEMU arguments table and editing controls as they caused user confusion around whether arguments were editable. Retain the "Export QEMU Command" debug button since the exported output and debug.log serve the actual debugging use case.


osy:
NACK, just removing it without a replacement is not helpful

PR #7632, opened 2026-02-26
vm(apple): implement USB passthrough for Apple Virtualization

### LINKED ISSUE #3778: Add USB sharing option to macOS guest (from macOS host)
With an e.g. Windows guest UTM offers sharing settings including for USB devices:

Screen Shot 2022-03-14 at 3 26 33 PM

But no such settings are available for a macOS guest:

Screen Shot 2022-03-14 at 3 26 05 PM

Can USB devices be shared with the new Virtualization framework? [Update: fwiw, I'm sure not seeing anything related to USB passthrough in Apple's <https://developer.apple.com/documentation/virtualization> overview documentation… ☹️ ]

(Sorry if this is a duplicate, I could not find one. This is in a somewhat similar vein as https://github.com/utmapp/UTM/issues/3567 — the new Virtualization.framework is still missing some features compared to the QEMU settings. File sharing is another such but I'm focusing this request on USB specifically since there's userspace options for transferring files, but not really any good workaround for passing through a USB device.)

osy:
This change implements USB device passthrough for macOS 15+ using the Virtualization framework private APIs.

Changes include:
- `UTMIOUSBHostManager`: Uses IOKit to dynamically manage connected USB devices. Instead of relying on private framework headers, it uses Objective-C reflection to safely instantiate `_VZIOUSBHostPassthroughDeviceConfiguration` and `_VZIOUSBHostPassthroughDevice`, and configures delegates for `VZUSBController`.
- `UTMIOUSBHostDevice`: Represents an IOKit USB device, conforming to `NSSecureCoding` and `NSCopying`. It safely reads properties such as location ID and port directly from the IORegistry.
- `UTMAppleVirtualMachine`: Now initializes `UTMIOUSBHostManager` and restores captured USB devices asynchronously before VM startup. State is archived directly into the registry.
- `VMDisplayAppleWindowController`: Added a new USB menu populated with current devices to allow interactively connecting and disconnecting USB devices on the fly.
- `UTMRegistryEntry`: Extended to support archiving and unarchiving of connected USB devices.

- [x] I have read the AI Contribution Guidelines and can attest that I have followed each item to the best of my ability.

### DIFF
--- a/Platform/macOS/Display/VMDisplayAppleWindowController.swift
+++ b/Platform/macOS/Display/VMDisplayAppleWindowController.swift
@@ -46,6 +46,9 @@ class VMDisplayAppleWindowController: VMDisplayWindowController {
     // MARK: - User preferences
     
     @Setting("SharePathAlertShown") private var isSharePathAlertShownPersistent: Bool = false
+    @Setting("NoUsbPrompt") private var isNoUsbPrompt: Bool = false
+    
+    private var allUsbDevices: [Any] = []
     
     override func windowDidLoad() {
         mainView!.translatesAutoresizingMaskIntoConstraints = false
@@ -94,11 +97,22 @@ class VMDisplayAppleWindowController: VMDisplayWindowController {
             setControl([.restart, .sharedFolder], isEnabled: false)
         }
         if #available(macOS 15, *) {
-            setControl(.drives, isEnabled: true)
+            setControl([.drives, .usb], isEnabled: true)
+        }
+        if #available(macOS 15, *), !isSecondary, let usbManager = appleVM.usbManager {
+            usbManager.delegate = self
         }
     }
     
     override func enterSuspended(isBusy busy: Bool) {
+        if #available(macOS 15, *), let usbManager = appleVM.usbManager {
+            usbManager.delegate = nil
+        }
+        if vm.state == .stopped {
+            if #available(macOS 15, *) {
+                allUsbDevices.removeAll()
+            }
+        }
         super.enterSuspended(isBusy: busy)
     }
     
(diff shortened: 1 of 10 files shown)


wjk:
> using the Virtualization framework private APIs

How do you expect this to pass Mac App Store review?

osy:
@wjk Well we're not doing anything different from what Parallels Desktop does. https://apps.apple.com/us/app/parallels-desktop/id1085114709

https://github.com/utmapp/UTM/issues/3778#issuecomment-3771461238

All I did was tell Gemini to look at what Parallels does and replicate it.

wjk:
😕 I suppose…

Parallels is a big company, meaning that Apple may be more inclined to look the other way concerning private APIs with their app as compared to UTM. And their private-API scanner must not be very sophisticated if it allows people to bypass it this easily.

On the other hand, you have had private CoreGraphics APIs referenced in the bridging header since 2021 per the blame, and the App Store lists versions back to 2022. I guess that’s OK with them? 😜

osy:
We expect all apps to be treated the same way as Apple and that all apps follow the same rules. If Apple has an issue with the use of these APIs, we will happily revert this change provided the same is asked of Parallels.

osy:
So it is with a heavy heart that I have to abandon this PR (at least until Apple makes the APIs public). The underlying assumption was that Parallels used the private APIs so we can too right? [Well, after extensive research, I can say that USB capture does not work in macOS guests in the App Store version.](https://github.com/utmapp/UTM/issues/3778#issuecomment-3974826558)

In order to get it working, it's not just a matter of calling the private APIs (which is quite easy to do as seen by the PR). You also need an entitlement `com.apple.security.temporary-exception.iokit-user-client-class` that Apple must manually review and grant (which I doubt they will). I did not commit the change where I added this entitlement because I had assumed that Parallels had a way of doing it without that entitlement but after a long and windy road, I've come to the conclusion that nobody at Parallels Desktop tests their products either (although TBF we are trying to build a testing infrastructure and we don't have hundreds of millions of dollars).

So I am going to leave this PR in draft for now. Until either 1) Apple makes this API public, or 2) we decide to maintain a non-sandboxed version of UTM macOS which is not difficult but comes with huge operational burden.

osy:
Messed up because I forgot my local main branch had the commit in place. I reverted the commit and opened a new PR #7635 to track this.

PR #7643, opened 2026-03-05
Added ability to get ip for AppleVirtualization machines

hellcastter:
Hello. I had a long-standing issue: I needed to fetch the IP address of my macOS machine, but it returned a 
```
Error from event: The operation couldn’t be completed. (OSStatus error -2700.)
Operation not supported by the backend.
``` 
I researched and found that IP will be stored at `/var/db/dhcpd_leases`. But I wasn't able to read it; it always said, "You have no permission". Then I changed my approach using ARP.

I created a solution that uses ARP to fetch the IP address of a VM running on AppleVirtualization and display it in the terminal.

Demo

This feature would allow us to implement `utmctl exec ...` via SSH (just need to think about password). Also, it would allow users to create some automations/servers/etc on their own.

Please consider adding this to the next build. Thank you.

I have read the AI Contribution Guidelines and can attest that I have followed each item to the best of my ability.

### DIFF
--- a/Scripting/UTMScriptingVirtualMachineImpl.swift
+++ b/Scripting/UTMScriptingVirtualMachineImpl.swift
@@ -278,9 +278,25 @@ class UTMScriptingVirtualMachineImpl: NSObject, UTMScriptable {
         }
     }
     
+    
     @objc func queryIp(_ command: NSScriptCommand) {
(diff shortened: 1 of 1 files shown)


hellcastter:
@osy, please can I have a review of my PR?

osy (approved):

PR #7631, opened 2026-02-25
tests: add UI test infrastructure for VM Wizard

sdrabb:
## Summary

- Add `UTMUITests` target wired into the macOS scheme so tests run via
  Xcode and `xcodebuild test -scheme macOS -destination 'platform=macOS'`
- Add `VMWizardViewTest` with a window close button smoke test
- Add `TESTING.md` guide covering how to run and extend the test suite
- Add `test-utm` CI job to `build.yml` that runs on every push/PR,
  parallel to `build-utm`, sharing the same sysroot cache

#2565

## Testing

Tested on: Mac mini M4, macOS 26.3 (Tahoe beta)

- `xcodebuild test -scheme macOS -destination 'platform=macOS'` passes locally
- this is the run I use to test it https://github.com/sdrabb/UTM/actions/runs/22518346473/job/65239516585#step:5:432

## Notes

Code is AI-assisted. All commits carry `Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>`.
I have read the AI Contribution Guidelines and can attest that I have followed each item to the best of my ability.

PR #7652, opened 2026-03-17
Add test plan

sdrabb:


sdrabb:
@osy This is my proposal for test plan, of course still in draft but I would like to ask you a first feedback before continuing in that direction, let e know what do you think about that

osy:
What is the difference between this and #7631? Have you addressed the comments from that review? It's not reasonable to ask for review for every 10-20 lines of code/documentation you write. Please look at how other PRs are reviewed and handled. Most come in complete while some are drafts that require some early input. Please do not create multiple drafts with minor changes. I believe the instructions were clear enough but if they are not clear, you can ask questions in the PR.

sdrabb:
In #7631 you ask for a human readable test plan without specifying any format, then I propose this asking for feedback.

Here I did a step back and I start creating a test plan and documenting how to contribute to it, like you ask.


Thanks

osy:
I said in that issue

> What I want to see (in human written form) is a test plan that covers everything that is important across all the different platforms (not just macOS but it is okay to start with just macOS).

This needs to be (minimum) hundreds of items.

sdrabb:
In my opinion we should build it incrementally.

What that matter as you mention is to have a plan and a clear definition on how people can start contribute to that.

Of course first iteration should include more than 2 cases but we cannot think to go 0 to 1 in one step, vision is what that matter here.

PR #7656, opened 2026-03-20
fix(ui): prevent crash when window size is zero

### LINKED ISSUE #7650: UTM crash if VM window resized to {0, 0, 0, 0}
**Describe the issue**  
While trying to resize VM windows using AppleScript, it occurred to me that the height of the titlebar might be included in the window's bounds, but I didn't know what that titlebar height was.

I ran this AppleScript:

```applescript
tell application "UTM"
	set bounds of window "vmname" to {0, 0, 0, 0}
end tell
```

The VM window disappeared, presumably having been made as small as possible and hidden under the top left corner of the menubar.

Then I ran this AppleScript:

```applescript
tell application "UTM"
	set b to bounds of window "vmname"
	set w to (item 3 of b) - (item 1 of b)
	set h to (item 4 of b) - (item 2 of b)
	return {w, h}
end tell
```

I was pleased to see the result:

```
{0, 38}
```

which told me the titlebar was 38 pixels tall. As I prepared to integrate this into my window resizing script, however, I noticed that UTM had crashed and thereby regrettably stopped my VMs which had been busy doing important work. It never occurred to me that adjusting UTM's UI using AppleScript could cause it to crash.

**Configuration**  
* UTM Version: 4.7.4
* macOS Version: 15.7
* Mac Chip (Intel, M1, ...): M1

**Crash log**  
[UTM-2026-03-17-082038.ips.txt](https://github.com/user-attachments/files/26047500/UTM-2026-03-17-082038.ips.txt)

**Debug log**  
N/A

**Upload VM**  
N/A

ty-yqs:
### Description
This PR addresses a critical issue reported in #7650 where setting a VM window's bounds to `{0, 0, 0, 0}` (e.g., via AppleScript or automation) causes the application to crash. This crash would forcefully terminate all running virtual machines, potentially leading to data loss.

The root cause was that the standard `NSWindow` allows setting a zero frame size, which subsequently causes failures in the layout or rendering pipeline (e.g., coordinate calculations resulting in NaN or division by zero).

### Changes
*   **New File**: `Platform/macOS/Display/VMDisplayWindow.swift`
    *   Implemented a custom `NSWindow` subclass named `VMDisplayWindow`.
    *   Added `sanitizeFrame(_:)` logic to enforce a minimum safe window size (400x300) and prevent invalid dimensions (width/height <= 0).
    *   Overrode `setFrame` methods to intercept all resize attempts and apply sanitized bounds.
    *   Added logging to warn developers/users when invalid window bounds are attempted.
*   **Modified**: `Platform/macOS/Display/Base.lproj/VMDisplayWindow.xib`
    *   Updated the window class from `NSWindow` to the custom `VMDisplayWindow` to activate the safe resizing logic.

### Testing Instructions
1.  Launch UTM and start a virtual machine.
2.  Open **Script Editor.app** and run the following AppleScript (replace `"vmname"` with your VM's window title):
    ```applescript
    tell application "UTM"
        set bounds of window "vmname" to {0, 0, 0, 0}
    end tell
    ```
3.  **Expected Result**:
    *   The app **does not crash**.
    *   The VM window resizes to the minimum allowed size (400x300).
    *   A warning is logged in the console: `"Attempted to set invalid window width..."`.

### Related Issue
Fixes #7650

### DIFF
--- a/Platform/macOS/Display/Base.lproj/VMDisplayWindow.xib
+++ b/Platform/macOS/Display/Base.lproj/VMDisplayWindow.xib
@@ -33,7 +33,7 @@
         </customObject>
         <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
-        <window title="UTM" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" animationBehavior="default" toolbarStyle="compact" id="QvC-M9-y7g">
+        <window title="UTM" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" animationBehavior="default" toolbarStyle="compact" id="QvC-M9-y7g" customClass="VMDisplayWindow" customModule="UTM" customModuleProvider="target">
             <windowStyleMask key="styleMask" titled="YES" closable="YES" miniaturizable="YES" resizable="YES"/>
             <windowCollectionBehavior key="collectionBehavior" fullScreenPrimary="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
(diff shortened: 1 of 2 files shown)


osy:
I think this is overkill for this issue. The GUI already prevents you from setting 0,0,0,0 so the only way to do this is through the scripting interface. The correct way to fix it is by validating the arguments from the scripting side (which may be difficult due to window.bounds being in a built in class). I do not want to modify the GUI and add all this code just to address this narrow issue that only happens in scripting. Also make sure to read the contributing guidelines.

PR #7657, opened 2026-03-21
docs: clarify UTM as Universal Turing Machine

anthumchris:
updated README to clarify the meaning of UTM. Explicitness is helpful to those who are new (like me) or don't know its meaning


osy:
I think it's more fun this way. You don't NEED to understand what it stands for and there's no official meaning beyond what makes sense to you.

PR #7682, opened 2026-04-11
Disable slide to delete on iOS

### LINKED ISSUE #3086: Extra safeguard step to prevent deleting VM
When trying out multiple different VM and operating systems I accidentally deleted one of the VMs-- nothing critical inside the VM but time to rebuild. Since it only takes a tap or swipe to "delete" a VM in UTM, perhaps could add a second question step before UTM deletes VM (e.g., Are you sure you want to delete this VM?, etc.).

jamesy0ung:
It is easy to slide to delete a VM on iOS and no confirmation is provided, so this has been removed via an ifdef for iOS. VMs can still be deleted by holding down on the name for the context menu and pressing "Delete", which provides an appropriate deletion confirmation.

Resolves https://github.com/utmapp/UTM/issues/3086

### DIFF
--- a/Platform/Shared/VMNavigationListView.swift
+++ b/Platform/Shared/VMNavigationListView.swift
@@ -67,8 +67,11 @@ struct VMNavigationListView: View {
                 }
             }
         }.onMove(perform: move)
+        
+        #if !os(iOS) // Disable slide to delete on iOS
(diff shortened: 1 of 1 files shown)


osy:
Fixed by 6df68ffe736aeb286f880168bd5652978b2009a1

PR #7693, opened 2026-04-27
Fix duplicate VM entries in sidebar and prevent data loss

### LINKED ISSUE #7533: [Bug] Duplicate VM entry appears in sidebar after launch; deleting one destroys the VM source file
Description:
I encountered a critical issue where a duplicate "ghost" entry of a running VM appears in the sidebar shortly after starting it. The list shows two identical VMs.

Thinking it was a UI glitch or a duplicate copy, I deleted one of them. However, this action deleted the actual .utm config/disk file from the file system. As a result, the remaining entry in the sidebar became unusable (cannot start or edit) because the source file was gone.

Configuration:

UTM Version: 4.7.4
macOS Version: macOS 26.2
Mac Chip: Apple Silicon M4

Crash Log:
No crash log, but the issue is reproducible in the UI.

Additional context:

The VM is an Ubuntu 24.04 ARM64 server.
The issue seems to stem from a state synchronization mismatch between the UI list and the actual running process/file.
Screenshot attached below showing the duplicate entries.

Image

DevSecTim:
## Summary

- Fix path comparison in `listRefresh()` to use filesystem identity (`fileResourceIdentifier`) instead of `standardizedFileURL` string comparison, preventing duplicate VM entries when security-scoped bookmark resolution returns a different URL form than `FileManager.contentsOfDirectory`
- Fix `VMData` equality operator to detect duplicates pointing to the same file on disk, even across different object instances
- Add safety check in `delete()` to skip file removal when another VM entry references the same `.utm` bundle, preventing data loss

## Context

On macOS 26.x, security-scoped bookmark resolution can produce URLs with different path forms (e.g., `/System/Volumes/Data/...` vs the regular mount point). This causes `listRefresh()` to fail to recognize an already-loaded VM when scanning the filesystem, creating a second entry for the same `.utm` file. When users delete what appears to be a "ghost" duplicate, the actual file is destroyed, breaking the remaining entry.

Full root cause analysis: https://github.com/utmapp/UTM/issues/7533#issuecomment-4326459068

Fixes #7533

## Test plan

- [ ] Verify VMs in the default storage directory appear exactly once in the sidebar
- [ ] Verify that after a bookmark-resolved URL differs from the filesystem path (can be tested by creating a symlink scenario), no duplicate appears
- [ ] Verify deleting a VM still removes the file when no duplicate references exist
- [ ] Verify that if a duplicate somehow exists, deleting one entry removes only the list entry, not the file

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/Platform/UTMData.swift
+++ b/Platform/UTMData.swift
@@ -158,10 +158,10 @@ enum AlertItem: Identifiable {
         }
         // now look for and add new VMs in default storage
         do {
-            let files = try fileManager.contentsOfDirectory(at: UTMData.defaultStorageUrl, includingPropertiesForKeys: [.isDirectoryKey], options: .skipsHiddenFiles)
+            let files = try fileManager.contentsOfDirectory(at: UTMData.defaultStorageUrl, includingPropertiesForKeys: [.isDirectoryKey, .fileResourceIdentifierKey], options: .skipsHiddenFiles)
             let newFiles = files.filter { newFile in
                 !list.contains { existingVM in
-                    existingVM.pathUrl.standardizedFileURL == newFile.standardizedFileURL
+                    Self.isSameFile(existingVM.pathUrl, as: newFile)
                 }
             }
             for file in newFiles {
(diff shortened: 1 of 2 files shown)


osy:
Fixed with 4879a465e6588b2d214f930b299348ff22fe4198

I didn't pick up this PR because as typical of Claude Code, there was a lot of unnecessary fluff that is unrelated to the issue and might cause more issues downstream.

saschpe:
Was this fix released already? I still get VM duplication on "clone" with UTM 4.7l.5 when using Apple Virtualization instead of QEMU.

Screenshot 2026-08-12 at 09 45 43

PR #7697, opened 2026-05-01
I fixed your godawful touch controls

ifilipis:
It's a complete rewrite of everything, but I did also preserve some old code for the time being

Let's start with issues:
- Old system had conflicts all over the place. Scroll would trigger pinch and taps all the time
- Gesture types were very limited
- Some recognizers were super flawed, like scroll that would only trigger if you drag it across half the screen
- I also discovered a few undocumented and unconfigurable gestures

At the times it was so bad that I just gave up and switched to Microsoft Remote Desktop, and a lot of gestures are inspired by it. They are quite typical to Microsoft tablets as well.

The list of changes is massive
- Old Touch mode conserved, new Multitouch mode added
- Tap / click still works the same way!
- 1 finger drag now does scroll and does not trigger erroneous clicks
- Added long press with default right click. Also added a visual indicator for when it is triggered.
- Added long press + drag for left clicking and dragging
- 2 finger tap now uses the location of first touch for mouse position. Much easier to aim this way
- 2 finger drag resolved conflicts with everything else and MUCH improved recognition. There's no more delay at all - it triggers instantly
- 2 finger swipe is also resolved correctly now and is fine-tunable. It took a lot of code to distinguish it from 2F drag
- 2 finger pinch also has a new resolver based on absolute finger travel instead of percentage to avoid erroneous triggers. This is what fixed it
- It is also 2 fingers only now. You can't do pinch with 3 or more fingers anymore
- Same set of fixes for 3 finger gestures + exposed 3 finger swipe in settings
- All gestures now persist! You can't trigger different gestures after any recognizer has fired.
- Added new types of mouse events, too. Middle drag, right drag, left drag + clicks
- Fixed a bug with gestures near the edges of the screen. There was a border where they were blocked, I even thought there's something wrong with my iPad
- I also played with defaults a bit. They are set to what MS RDP does at the moment
- Nothing in the driver was changed. All fixes are in the UI

Hoping to see it in the upcoming release, because otherwise, the UX is horrible and made me dust off a MacBook, install macOS updates and Xcode for this sole purpose

### DIFF
--- a/Platform/iOS/Display/VMDisplayMetalViewController+Private.h
+++ b/Platform/iOS/Display/VMDisplayMetalViewController+Private.h
@@ -44,18 +44,45 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable) VMScroll *scroll;
 
 // Gestures
-@property (nonatomic, nullable) UISwipeGestureRecognizer *swipeUp;
-@property (nonatomic, nullable) UISwipeGestureRecognizer *swipeDown;
 @property (nonatomic, nullable) UISwipeGestureRecognizer *swipeScrollUp;
 @property (nonatomic, nullable) UISwipeGestureRecognizer *swipeScrollDown;
 @property (nonatomic, nullable) UIPanGestureRecognizer *pan;
 @property (nonatomic, nullable) UIPanGestureRecognizer *twoPan;
 @property (nonatomic, nullable) UIPanGestureRecognizer *threePan;
+@property (nonatomic, nullable) UIPinchGestureRecognizer *pinch;
 @property (nonatomic, nullable) UITapGestureRecognizer *tap;
 @property (nonatomic, nullable) UITapGestureRecognizer *tapPencil;
 @property (nonatomic, nullable) UITapGestureRecognizer *twoTap;
+@property (nonatomic, nullable) UITapGestureRecognizer *threeTap;
 @property (nonatomic, nullable) UILongPressGestureRecognizer *longPress;
-@property (nonatomic, nullable) UIPinchGestureRecognizer *pinch;
+@property (nonatomic, nullable) UITouch *multitouchPrimaryTouch;
+@property (nonatomic) BOOL multitouchTwoPanConsumed;
+@property (nonatomic) BOOL multitouchThreePanConsumed;
+@property (nonatomic) BOOL multitouchTwoPanActionStarted;
+@property (nonatomic) BOOL multitouchThreePanActionStarted;
+@property (nonatomic) BOOL multitouchTwoSwipeDecided;
+@property (nonatomic) BOOL multitouchThreeSwipeDecided;
+@property (nonatomic) BOOL multitouchTwoSwipeCandidate;
+@property (nonatomic) BOOL multitouchThreeSwipeCandidate;
+@property (nonatomic) CGPoint multitouchTwoPanLastVelocity;
+@property (nonatomic) CGPoint multitouchThreePanLastVelocity;
+@property (nonatomic) NSTimeInterval multitouchTwoPanLastTime;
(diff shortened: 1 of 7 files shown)


osy:
Please confirm you have read and followed all of the contribution guidelines: https://github.com/utmapp/UTM/blob/main/CONTRIBUTING.md

keesverruijt:
@ifilipis your PR might the most wonderful PR ever, but why the attitude? Can't you keep it civil please? Some modesty, kindness and politeness goes a long way.

GBirkel:
Calling software godawful and UX horrible is certainly harsh and a disservice to the effort of the authors, but I don't know if it crosses the line into uncivil...  When UX has deep flaws, the frustration it causes can be really intense. :D

But it's moot.  By the end of this year it's likely that _every_ PR summary we read, everywhere on this site, will be auto-generated by LLM, and they will all sound the same, and we will not be reading words that came out of actual humans in PRs ever again.  Indicators of politeness and positive attitude will become so ubiquitous they will stop having their intended effect and instead become useless bulk in the language and in due time they will be also scraped out, and PRs will consist entirely of terse bulleted lists (which is what this PR description mostly is, coincidentally)

End of an era, folks!

ifilipis:
@GBirkel Thank you for calling my hand-written list AI-generated. Your paranoia is definitely accurate. And honestly? Enjoy the godawful UX, too!

keesverruijt:
@ifilipis You need to learn to read more carefully. @GBirkel was actually defending you somewhat IMO. He did certainly not suggest that you used AI. He suggested that soon everyone will use AI for all PRs. I don't agree with him, human only texts may become rare voices but I don't think _everything_ will be AI, unless Github gives them an extra leg up like "PRs only allowed by AIs that follow the submission rules of the repository."

Ickerday:
So @ifilipis backs down with the attitude, adjusts the PR title and description, someone reviews the code and we all win, right?

osy:
@Ickerday no as I've stated above

> Please confirm you have read and followed all of the contribution guidelines: https://github.com/utmapp/UTM/blob/main/CONTRIBUTING.md

There's nothing about attitude or PR title there. I don't care if you're an arsehole as long as you can follow instructions.

osy:
Superseded by #7893.

The list of problems in this PR's description was useful: most of them were real, and #7893 goes through all twenty points one by one and says what was done about each. The changes here can't be merged as they are, though, because they don't follow the [contribution guidelines](https://github.com/utmapp/UTM/blob/main/CONTRIBUTING.md):

- It is a single 900-line commit covering many unrelated fixes and features, where the guidelines ask for one well-defined change per PR and commits titled `component: short description` that explain why.
- Despite "Old Touch mode conserved", it changes behaviour for existing users: pinch to zoom and the three finger keyboard swipe are gated behind new settings that default to Disabled, the two finger "Mouse Wheel (per swipe)" stops working in the old modes, and several gesture defaults are changed for everyone who never opened Settings.
- It adds four new settings, against "prefer simplicity over options".
- There is no record of the required testing (device and OS version, and for bug fixes reproducing the bug before and confirming the fix after), and the confirmation requested above was never given.

#7893 is a rewrite that follows the guidelines and does the required testing: one commit per issue, each reproduced and verified before and after on iOS 17, 18, 26 and 27 simulators, then tested on a device. It also fixes some latent bugs that this PR worked around or didn't cover (the long press waiting on the tap recognizer in every mode, every pan waiting up to half a second on the keyboard swipes, a cancelled drag leaving the button held down in the guest, the pinch taking three finger gestures on iOS 17, and double tap in the new mode), and the code is considerably smaller: two pieces of state instead of about thirty, the existing touch modes left untouched, and the public SwiftUI modifier for edge gestures instead of patching `UIViewController`.

Thank you for identifying these issues. Closing in favour of #7893.

PR #7816, opened 2026-08-08
Fix bug where `listRefresh()` would not deduplicate already existing VMs in list

avighnac:
While I was using UTM, I noticed that sometimes, after creating a VM, it would create two entries for that VM. On trying to delete one of the entries, the whole VM would be deleted though. 

I'm not sure why the duplicate entries are being created but after looking at the code for some time, I found this function which claims to remove duplicate entries.

The thing is, however, it only seemed to remove 'newly loaded' duplicates. I added some code at the very beginning that makes it remove already existing duplicates too.

I tested this myself (Macbook Air M2, macOS Tahoe 26.5) and a Windows VM that appeared twice on the regular UTM appears only once with this patch.

### DIFF
--- a/Platform/UTMData.swift
+++ b/Platform/UTMData.swift
@@ -144,12 +144,18 @@ enum AlertItem: Identifiable {
     ///
     /// This removes stale entries (deleted/not accessible) and duplicate entries
     func listRefresh() async {
+        // deduplicate VM list
+        var list: [VMData] = [];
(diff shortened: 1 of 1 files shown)


avighnac:
Bump

osy:
Does this bug still happen on v5.0.5? There's been some work there if I remember correctly.

PR #7840, opened 2026-08-29
input(macOS): preserve Caps Lock key releases

### LINKED ISSUE #7838: Caps Lock remains held in guest after key is released
Describe the issue

When Caps Lock is treated as a key is enabled, Caps Lock can be used as a held modifier in a guest, but releasing the physical Caps Lock key does not appear to release it in the guest.

This is reliably reproducible with NVDA in a Windows 11 ARM QEMU guest when Caps Lock is configured as the NVDA modifier. After using a Caps Lock-based NVDA command and releasing the key, the next key is still interpreted as though Caps Lock is held. Pressing Caps Lock once more clears the state.

This is particularly noticeable with screen readers because they use Caps Lock as a held modifier rather than only as a toggle. I have not confirmed whether other software that depends on Caps Lock key-down/key-up events is affected.

Related to #4330, which was closed after the Caps Lock is treated as a key preference was available as a workaround for screen-reader use.

Steps to reproduce

1. In UTM settings, enable Caps Lock is treated as a key.
2. Start a Windows 11 ARM QEMU guest.
3. In NVDA, configure Caps Lock as the NVDA modifier.
4. Hold Caps Lock and press an NVDA command, for example Caps Lock+N, then release both keys.
5. Press a normal key such as T.

Expected behavior

After Caps Lock is physically released, the next key should be handled normally. In the example above, T should be a normal T.

Actual behavior

The next key is still interpreted as though Caps Lock is being held. In the example above, NVDA handles T as NVDA+T. Pressing and releasing Caps Lock again clears the stuck modifier state.

Configuration

* UTM Version: 4.7.5 (118)
* macOS Version: 26.6.2
* Mac Chip: Apple M5 Max
* Guest: Windows 11 ARM, QEMU

Crash log

Not applicable; UTM does not crash.

Debug log

Not attached. The issue is reliably reproducible as described above; I can provide a debug log if it would help diagnose the input event handling.

Upload VM

Not attached. The issue is reproducible in a standard Windows 11 ARM QEMU guest.

lprazdnik:
## Summary

- Preserve Caps Lock key-down and key-up events when “Caps Lock is treated as a key” is enabled.
- Temporarily remap Caps Lock while guest input is captured and restore the host mapping afterward.
- Prevent Caps Lock-based guest modifiers, such as NVDA commands, from remaining stuck after the key is released.

Resolves #7838

## Testing

Tested by a human on a MacBook Pro with Apple M5 Max running macOS 26.6.2, using Windows 11 ARM as a QEMU guest with NVDA.

Verified repeated commands while Caps Lock remained held, normal input after releasing Caps Lock, and restoration of the host keyboard mapping after releasing VM input.

The author acknowledges that this change has been tested and/or reviewed by a human in accordance with UTM's AI contribution guidelines.

### DIFF
--- a/Platform/macOS/AppDelegate.swift
+++ b/Platform/macOS/AppDelegate.swift
@@ -152,6 +152,8 @@
     func applicationWillTerminate(_ notification: Notification) {
         /// Synchronize registry
         UTMRegistry.shared.sync()
+        /// Give the host its Caps Lock back if a VM still has input captured
+        CapsLockRemapper.shared.restore()
         /// Clean up caches
         let fileManager = FileManager.default
         guard let cacheUrl = fileManager.urls(for: .cachesDirectory, in: .userDomainMask).first else {
@@ -169,6 +171,7 @@
     }
     
     func applicationDidFinishLaunching(_ notification: Notification) {
+        CapsLockRemapper.shared.recoverAtLaunch()
         if isDockIconHidden {
             NSApp.setActivationPolicy(.accessory)
         }
--- a/Platform/macOS/Display/CapsLockRemapper.swift
+++ b/Platform/macOS/Display/CapsLockRemapper.swift
@@ -1,0 +1,108 @@
+//
+// Copyright © 2026 UTM contributors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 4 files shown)


osy:
@lprazdnik I've re-written the commit, can you re-test?

lprazdnik:
I tested the latest commit with Windows 11 ARM and NVDA. Caps Lock works as expected, including repeated commands while held and normal input after release.

Thanks!

PR #7853, opened 2026-09-05
spice-gtk: fix SPICE deadlock under memory pressure

evan314159:
CoreAudio AudioOutputUnitStop() hangs under high memory load (e.g. concurrent LLM usage).  GStreamer calls AudioOutputUnitStop() in the main context thread, causing all SPICE services including video and input to hang while the VM is still running.

Short-term fix (this PR): run GStreamer state changes on a separate, serialised worker thread per stream so the hang does not block other SPICE services.

Long-term fix?: this appears to require either a fix from Apple or for audio to run in a separate process. Once the connection to CoreAudio is hung, it does not seem to be recoverable by the same process.  FB24720735 opened with Apple.

To reproduce:
* Start a Linux VM using QEMU, with a graphical desktop and optionally a serial console, and play a video.
* Start a local LLM app like LM Studio and load a large model taking almost all memory (e.g. 17GB of a 24GB system) causing memory pressure but not complete failure.
* VM video, audio and input will hang.  The serial console will remain available and there will be no console or dmesg errors.

Developed with AI but human tested.  Before the change, all video, audio, and input hung within a few minutes of running a local LLM, and live process sampling (`sample(1)`) found that the main context was hung in AudioOutputUnitStop().  Post-change, audio still hangs but video and input are unaffected, and only the worker is blocked.  Test system: macOS 27.0 latest developer beta (26A5425a), UTM 5.0.5 beta.

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/patches/spice-gtk-0.42.patch
+++ b/patches/spice-gtk-0.42.patch
@@ -1,7 +1,7 @@
 From 07ba2d801b4a03125dee3f9d5f4a13cad8d62008 Mon Sep 17 00:00:00 2001
 From: osy <osy@turing.llc>
 Date: Fri, 4 Mar 2022 16:35:26 -0800
-Subject: [PATCH 1/2] spice-util: support for non-default GMainContext
+Subject: [PATCH 1/5] spice-util: support for non-default GMainContext
 
 When spice-gtk is used in an application with its own GMainContext, the
 wrong context will be used leading to various issues.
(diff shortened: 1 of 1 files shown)


evan314159:
Also confirming I read https://github.com/utmapp/UTM/blob/main/CONTRIBUTING.md and as far as I can tell followed it.  I used /utm-review but not /utm-submit.

evan314159:
The failed check looks like a build timeout and not an issue with the PR.

Issue #5377, opened 2023-06-08
utmctl list gives error: OSStatus error -1743


muzzah:
**Describe the issue**  
This has been reported in a closed issue so Im creating a new ticket for this as I think its an important fix  thats needed.

When using utmctl over ssh, the list command returns an error. This seems to be a gap in capability as I cannot now remotely manage the VMs. 

#5001 Is the original issue where this was reported. 

```
[9:41:21] universe:bin $ sudo utmctl list
Error from event: The operation couldn’t be completed. (OSStatus error -600.)
````

**Configuration**  
* UTM Version: 4.2.5
* macOS Version: 13.4
* Mac Chip (Intel, M1, ...): M1

**Comment by muzzah** [comment]
As a side note, I looked into the `config.plist` under the `~Library/Containers/com.utmapp.UTM/Data/Documents/[VM NAME] and even if I use the ID or Name of the VM with `utmctl` it still fails. 

I used the key located under UUID below, if this is not correct, can you advise which UUID I should be using? Why does it not work with the name field?

```
<dict>
		<key>Icon</key>
		<string>linux</string>
		<key>IconCustom</key>
		<false/>
		<key>Name</key>
		<string>Dev Server</string>
		<key>UUID</key>
		<string>8EB11628-544C-44BF-9EDE-4A65A616BEE2</string>
	</dict>
```

antekone:
I have the same problem, I'm not able to manipulate VMs by using `utmctl` from an SSH session, even if I'm logged in to GUI in another session.

`utmctl` from a Terminal app from GUI works fine.

osy:
This is unfortunately a limitation of how apple scripting bridge works. You must have the GUI launched or launchable in an interactive instance. I don’t think it works with SSH.

nre-ableton:
> This is unfortunately a limitation of how apple scripting bridge works. You must have the GUI launched or launchable in an interactive instance. I don’t think it works with SSH.

This makes sense, thanks for the explanation. Is there no way to work around this? For example, by requiring the same user to be logged-in and having the UTM app running?

I was really hoping that our company could use UTM for Mac VMs, but unfortunately as we use Ansible for VM deployments (which runs over SSH), this limitation is a dealbreaker for us. However, our Ansible playbooks use `osascript` in other places (for example, `osascript -e 'tell application "System Events" to keystroke RETURN` to accept system dialog popups), so it seems like there _should_ theoretically be a way to get this working. :thinking:

[...]

PR #7855, opened 2026-09-06
utmctl: add native local control interface

synrest:
## Summary

Adds a native local control path for utmctl on macOS, replacing most ScriptingBridge-based VM control with a structured local App Group Unix-domain socket owned by UTM.app.

UTM.app remains the canonical authority through UTMData/VMData/backend services.

## Motivation

The existing ScriptingBridge path provides weakly structured control and error behavior, and its framework-mediated application-launch semantics are not a reliable foundation for local VM administration. This change introduces a native local control interface rather than extending those paths further.

Upstream issue #5377 is included as historical context for limitations in the ScriptingBridge-based approach; this PR does not claim to resolve or close it.

## Architecture

utmctl
  ↓
App Group Unix-domain socket
  ↓
UTM.app
  ↓
UTMData / VMData / canonical backend services

The protocol uses Codable bounded request/response messages, structured errors, exact UUID/exact-name VM resolution, per-VM lifecycle serialization, and SIGPIPE-safe transport.

There is no TCP control endpoint, arbitrary shell transport, or implicit graceful-stop → force-stop escalation.

## Authority policy

Commands use centralized `local`, `wakeAllowed`, or `requireRunning` policies.

`wakeAllowed` commands may bootstrap the local authority. `requireRunning` commands fail cleanly when the authority is absent. CLI-originated bootstrap is hidden and nonactivating.

## Presentation semantics

- Graphical configuration → graphical presentation
- Terminal configuration → terminal presentation
- Headless configuration → headless session

`utmctl start` does not redefine VM presentation.

## Migrated CLI surface

- list
- status
- start
- stop
- force-stop
- suspend
- resume
- clone
- delete
- ip-address
- snapshot create/list/restore/delete
- usb list/connect/disconnect
- exec
- file pull/push

Version is handled locally from bundle metadata. File transfer uses bounded offset-based chunked RPC.

## Deferred

`attach` remains intentionally unmigrated because it requires a proper streaming serial-session abstraction. The current command is already incomplete, and long-lived bidirectional streaming should not be forced into the bounded request/response protocol.

## Human validation

Human-tested on macOS 15.7.9, x86_64.

- Canonical archive/package build: PASS
- Signing and CLI gates: PASS
- Fresh reconstructed artifact:
  - list: PASS
  - status by exact VM name: PASS
  - start by exact VM name: PASS
  - stop by exact VM name: PASS

Broader validation covered graphical/headless lifecycle, suspend/resume, clone/delete, IP querying, snapshots, USB routing, guest exec, and chunked file transfer.


osy:
This overturns the design decision which is to reduce the number of separate backends which can drift in feature and make adding new features painful. I would rather fix any existing issues than rewriting a bunch of code with duplicate features.

synrest:
Thanks, that makes sense. The intent was to address the limitations we encountered with the existing control path, particularly around non-interactive CLI use, but I understand the concern about maintaining two backends and feature drift. I'll take another look at whether those issues can be addressed within the existing backend instead.

Issue #3812, opened 2022-03-24
Tabs like VMware?


DJZSwag:
I'm not a fan of having multiple windows on my desktop, and a lot of people don't want their desktop to be cluttered with windows, too. Is there a possible way of adding tabs like VMware? If so, could you make it in the next big release of UTM?
![Workstation-1](https://user-images.githubusercontent.com/87071118/159917027-98a77a5a-84e8-42a2-80d8-09dba70297ba.jpeg)

DJZSwag:
I've realized something: If you add tabs, then you would have to redesign the entire home UI. That's why think this should be added in a major update, like 4.0.

DJZSwag:
It's a good Idea to put an option to disable Tabs. Might This is good alongside redesigned preferences.

adespoton:
Anyone ever used [MobaXTerm](https://mobaxterm.mobatek.net/)?  Windows tool for managing shells, RDP/VNC sessions, etc. that has horizontal and vertical tabs, and allows you to set on a per-bookmark basis how those should open -- in a tab (default), windowed or full-screen (or "integrated desktop" via X server).  You can switch any window between the 

I've always really liked the interface, and found it to be more Mac-like than many macOS UIs around these days.

DJZSwag:
Looks cool! Too bad my windows laptop is not working well.

conath:
It might be possible to use the macOS document-based application API to take advantage of the system’s multi-window & tab functionality.

jrjsmrtn:
I recently discovered [Royal TSX](https://www.royalapps.com/ts/mac/features), a remote connection management macOS application, supporting VNC, RDP, SSH, Web, etc. 

It has a very pleasant and efficient UI, with a hierarchical Navigation panel on the left, a splittable, tabbed pane on the right with an always present "Overview" tab, showing miniatures of active connections, and tabs for every active connections.

This could give you some inspiration... ;-)

saschpe:
I would love a tabbed interface as well. Especially when working with multiple VMs I'd assume switching tabs (given proper shortcuts) is much faster than selecting windows.

PR #7859, opened 2026-09-11
macOS: option to open VM windows as tabs of the main window

LayerTM:
Addresses #3812.

Many people run one or two VMs and would rather keep the VM screen inside UTM's main window than manage a separate window for each VM. macOS already provides this through native window tabbing, but UTM turns tabbing off app-wide (`NSWindow.allowsAutomaticWindowTabbing = false` in `ContentView`).

This adds a setting, **Settings → Application → "Open virtual machine windows as tabs of the main window"**, off by default, so nothing changes for anyone who does not enable it.

When it is on:
- a VM window opened from the library, or through AppleScript / `utmctl start`, is attached as a tab of the library window (`addTabbedWindow(_:ordered:)`), since both paths go through `UTMData.run(vm:)`;
- automatic window tabbing is no longer disabled at launch (read at launch, so this part takes effect after a relaunch);
- if the library window is closed, the VM window opens on its own as it does today.

Changes are limited to three files: `Platform/Shared/ContentView.swift` (one line), `Platform/macOS/SettingsView.swift` (the toggle), `Platform/macOS/UTMDataExtension.swift` (the attach step).

Tested on macOS 26 (Apple silicon) with an unsigned build of this branch and a QEMU VM (aarch64, display on):
- start from the library → the VM screen opens as a tab of the main window;
- `utmctl start <vm>` with UTM already open → same, no separate window;
- three starts in total (library button, AppleScript/`utmctl` twice), no crashes.

With the setting off, the attach step returns before doing anything and `allowsAutomaticWindowTabbing` is set to `false` exactly as before (read from the code; not separately exercised).

The new string is not localized yet.


LayerTM:
Friendly ping on this one. The Build workflow for this PR has never run — it is still sitting at "action required", waiting for a maintainer to approve the run for a first-time contributor, so there is no CI signal on it either way.

GitHub still reports the branch as merging cleanly on current main. If it would help, I am happy to rebase it, split it up, or change the approach — just say which. If the feature is not wanted, closing it is a fine outcome too; I would rather know than leave it open.

Issue #7418, opened 2025-09-17
Support for VZFileHandleNetworkDeviceAttachment & VZVmnetNetworkDeviceAttachment


ctison:
It would be awesome to support:
- [VZVmnetNetworkDeviceAttachment](https://developer.apple.com/documentation/virtualization/vzvmnetnetworkdeviceattachment) (`macOS 26.0+`)
- [VZFileHandleNetworkDeviceAttachment](https://developer.apple.com/documentation/virtualization/vzfilehandlenetworkdeviceattachment) (`macOS 11.0+`)

Supporting these would enable advanced network architectures, such as running a Linux routing VM with:
  - A `VZFileHandleNetworkDeviceAttachment`, allowing packets to be read and written from the host, and routed through a VPN. This is especially useful when LAN connections are blocked (Docker Desktop uses a similar approach with [VPNKit](https://github.com/moby/vpnkit)[^1]).
  - A `VZVmnetNetworkDeviceAttachment`, which can also be mounted on other VMs to allow them to access the internet via the Linux routing VM.

This would enable cleaner solutions compared to existing workarounds:
- https://amodm.com/blog/2024/07/03/running-a-linux-router-on-macos

Note: #7380 did some work for `VZVmnetNetworkDeviceAttachment`.

[^1]: https://docs.docker.com/desktop/features/networking/#vpn-passthrough

nirs:
For [VZVmnetNetworkDeviceAttachment](https://developer.apple.com/documentation/virtualization/vzvmnetnetworkdeviceattachment) (macOS 26.0+) we can integrate with [vmnet-broker](https://github.com/nirs/vmnet-broker).

How it works:
- User select vment shared network or vment host network
- VM launcher acquire a network from the broker via VmnetBroker Swift library and create a vment_network_create_with_serialization
- The network will be available while the VM launcher process is running
- No additional code is needed, the broker manage network lifetime.
- example: https://github.com/nirs/vmnet-broker/blob/e536fe3f391c3aa53cd0a1382badad74b3e2b5c7/swift/Sources/test/main.swift#L230

Note: the broker is not released yet.

For [VZFileHandleNetworkDeviceAttachment](https://developer.apple.com/documentation/virtualization/vzfilehandlenetworkdeviceattachment) (macOS 11.0+) we can integrate with [vmnet-helper](https://github.com/nirs/vmnet-helper).

How it works:
- User select vment shared network or vment host network
- The VM launcher process creates a socketpair:
[...]

PR #7865, opened 2026-09-14
apple: let virtual machines connect to each other

### LINKED ISSUE #6975: Allow VM to VM connections
What's the best to connect to a VM from another VM in UTM apple virtualization mode? Please consider supporting like lima's [user-v2](https://lima-vm.io/docs/config/network/#lima-user-v2-network) network or VMware fusion's shared network.

### LINKED ISSUE #7480: host-only networks for non-emulated VMs w/ "Apple Virtualization"
Hello,

not sure whether this is a documentation question or feature request.

I try to implement multiple network interfaces on virtual machines created via UTM 4.7.4 on Apple Silicon M3 and MacOS 15.7.1, while the VM is created with "Apple Virtualization", i.e. without qemu.

Accordingly, the only available options for the NIC types seem to be "Shared" and "Bridged".
Is it possible, to implement multiple NICs, each of which are in different subnets, similar to VirtualBox "Internal Networking"?
The documentation is a bit thin here...

From the experience i made, the core problem seems to be that multiple "Bridged" network interfaces can be implemented per VM, but they all connect to the same bridge bridge101, which kind of defeats the purpose, if i did not miss s.th.

Also, "bridge101" seems to autoconnect to the respective currently active NIC of the host, which makes networking of virtual deployments somewhat hard to predict, because the IP address range changes with any different environment the hypervisor is connected to?

Best

lovon-spec:
## Summary

Add two Apple Virtualization networking modes on macOS 26 and later using `VZVmnetNetworkDeviceAttachment`:

- **Shared Network (VM to VM)** puts Apple VMs on the same vmnet shared network so they can communicate with each other while retaining host and internet access.
- **Host Only** puts Apple VMs on the same host-only vmnet network so they can communicate with each other and the Mac without internet routing.

The vmnet manager weakly caches one attachment per mode while Virtualization objects own its actual lifetime. Configuration validation stays side-effect-free, so saving a VM does not create a live vmnet network. The new modes also round-trip through UTM's scripting interface and are hidden on older macOS versions.

Resolves #6975.
Related to #7418 and #7380.

## Testing

I tested this change on **MacBook Pro with Apple M5 Pro running macOS 26.4.1**.

- **Shared Network (VM to VM):** I started two Apple Virtualization VMs, checked their guest IP configuration, and confirmed direct guest-to-guest SSH connectivity.
- **Host Only:** I confirmed guest-to-guest connectivity, connectivity to the host, and verified that internet access was unavailable as intended.

I confirm that this change has been tested and reviewed in accordance with UTM's AI contribution guidelines.

Additional automated application-level verification covered pause/resume, first-user stop/restart, final network release, saved-state restore, concurrent configuration saves, scripting round-trips, network-creation errors without NAT fallback, and the post-start external-drive failure lifetime case.


ttc0419:
@osy Could you take a look? It's the last missing piece for many users.

osy:
Patience, there's like a dozen things I have to review.

osy:
@ttc0419 I've made some changes and synced with the latest HEAD. Can you re-test?

lovon-spec:
@osy Thank you for the review and the cleanup here! I retested `155ad876` on both macOS 26.4.1 and 27 and found one issue with the preferred addresses.

The short version is that vmnet treats the address we pass as the host/router address for that `/24`, so using `192.168.96.0` / `192.168.160.0` makes the VM interface fail later during startup. Using `.1` instead keeps the same subnets but gives the host a valid address.

I confirmed the two-line `.0` → `.1` change with a source-built UTM on macOS 27: guest/host connectivity works in both modes, NAT gets internet access, and Host Only stays isolated. The fallback doesn’t catch the current failure because network creation itself succeeds; the interface only fails afterward.
Happy to fold that fix into the existing commit if you want.

Otherwise, the networking and lifecycle paths I retested looked good.

lovon-spec:
PS: I went ahead and pushed the two-line `.0` → `.1` fix as `47eadda150b7`.

PR #7885, opened 2026-09-19
display: fill the area beside the camera housing in full screen

### LINKED ISSUE #6183: "true" fullscreen on macbook with notch
Great app!

Currently, on a macbook with notch, fullscreen mode is displayed with black margin at the top:
now
It works the same in most apps, and it's reasonable by default because otherwise some content may be hidden behind the notch.

But for some use-cases, in opposite, it might be fancy to cover whole screen and let a little content go behind the notch:

https://github.com/utmapp/UTM/assets/18294052/7a8f5abd-a171-45c6-9586-a9d66533f7e9

Here is what I found (from user side):
https://www.reddit.com/r/MacOS/comments/qveb0u/force_true_fullscreen_on_macbook_with_notch/
https://notes.alinpanaitiu.com/Fullscreen%20apps%20above%20the%20MacBook%20notch

osy:
On a Mac with a camera housing, a full screen VM loses the strip of screen beside it: AppKit insets the full screen frame and the window server covers the inset area with an opaque menu bar. This adds an opt-in preference (Settings → Display, macOS 27+) that gives that area to the guest display instead.

The window keeps using native full screen — it is not a borderless window, so macOS still treats the VM as a full screen app, and Mission Control, Spaces and Split View behave as before.

The menu bar is normally revealed as soon as the pointer is anywhere over that strip, which would make the area unusable for the guest. With the option on it is only revealed when the pointer reaches the very top edge of the screen; until then the space hides the menu bar (and with it the Dock), and clicks in the strip go to the VM. Once revealed, the menu bar and toolbar look and behave as they always did.

Everything this relies on is private AppKit and SkyLight API. None of it is used or hooked until the option is enabled and a window enters full screen, so with the option off the window behaves exactly as before. The symbols are looked up at runtime by name and checked against the signatures we expect, so nothing private is linked or compiled into the binary, and if any of them is missing or different the window falls back to normal full screen. It also stands down when tiled in Split View, on a screen without a camera housing, and when the menu bar is set to always be visible in full screen. Terminal windows do not opt in.

Resolves #6183

**Testing:** Tested by a human on **macOS 27.0, MacBook Air M3**. The author acknowledges that this change has been tested and/or reviewed by a human in accordance with UTM's AI contribution guidelines.

Tested with the option on and off, on QEMU (fixed resolution UEFI guest, and Windows 11 with SPICE tools which changes its resolution) and on an Apple backend VM: entering and leaving full screen, Split View, switching spaces away and back, revealing and hiding the menu bar, and clicking in the strip.


kmatheussen:
Thank you.

